PR #20, opened 2021-11-17
Updating Sass version to be able to use Math.div

domonkosgabor:


domonkosgabor:
Sass version updated to 1.43.4, the front-end of the solution tested with the new version. There is no breaking change regarding to the [official documentation](https://sass-lang.com/documentation/breaking-changes/slash-div#automatic-migration) because their is nothing to migrate.

PR #41, opened 2022-01-22
OSOE-49: ⭐ Replace Windows-only code in UI Testing Toolbox and Fix Lombiq.Analyzers on Linux

sarahelsaig:
OSOE-49


Piedone:
Merge from dev.

sarahelsaig:
ok, merged

Piedone:
And fix the build errors too :).

.github/workflows/analyze_and_test.yml
```diff
@@ -0,0 +1,50 @@
+name: Pull Request - Code Analysis and Unit Tests
+on:
+  pull_request:
+    branches: [ dev ]
```
Piedone: We sometimes open longer-living feature branches with multiple PRs to them, so we don't need this restriction.

.github/workflows/analyze_and_test.yml
```diff
@@ -0,0 +1,50 @@
+name: Pull Request - Code Analysis and Unit Tests
```
Piedone:
```suggestion
name: Pull Request - Build and Test
```

.github/workflows/analyze_and_test.yml
```diff
@@ -0,0 +1,50 @@
...
+on:
+  pull_request:
+    branches: [ dev ]
+env:
+  DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+  DOTNET_CLI_TELEMETRY_OPTOUT: true
+jobs:
+  build_test:
+    runs-on: ubuntu-latest
+    name: Build & Test
```
Piedone: Just to be in line with TC:

```suggestion
    name: Build and Test
```

.github/workflows/analyze_and_test.yml
```diff
@@ -0,0 +1,51 @@
...
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 6.0.*
+    - name: Submodule Init
+      run: |
+        git submodule init && git submodule update
+    - name: Static Code Analysis
+      run: |
+        npm install pnpm -g &&
+        dotnet build src/Utilities/Lombiq.Npm.Targets/Lombiq.Npm.Targets.csproj
```
Piedone: Wouldn't project references to NPM Targets, just as to Gulp Extensions, fix this too?

.github/workflows/analyze_and_test.yml
```diff
@@ -0,0 +1,51 @@
...
+        dotnet build Lombiq.OSOCE.sln --no-incremental -warnaserror /p:TreatWarningsAsErrors=true /p:RunAnalyzersDuringBuild=true -nologo -consoleLoggerParameters:NoSummary -verbosity:quiet
```
Piedone: `--no-incremental` shouldn't be needed on a single-use build agent.

Also, better to build like we'd run it in prod: `--configuration Release`. I don't really understand why there's a second Release build after this.

And why quiet?

.github/workflows/analyze_and_test.yml
```diff
@@ -0,0 +1,51 @@
...
+    - name: Build
+      run: |
+        dotnet build --configuration Release
+    - name: Set up SQL Server via Docker
```
Piedone: Wow, nice!

.github/TestConfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "Lombiq_Tests_UI": {
```
Piedone, line 2: It would be a bit nicer to keep these configs in environment variables, see `TestConfigurationManager` (you can make more code configs configurable from the env, just add more `TestConfigurationManager` calls with defaults keeping the current defaults).

.github/workflows/analyze_and_test.yml
```diff
@@ -0,0 +1,51 @@
...
+        docker pull mcr.microsoft.com/mssql/server
+        docker volume create --driver local -o o=bind -o type=none -o device=~/.local/docker/mssql/data mssql-data
+        docker run --name sql2019 -e 'ACCEPT_EULA=Y' -e 'SA_PASSWORD=yourStrong(!)Password' -v mssql-data:/data -p 1433:1433 -d 'mcr.microsoft.com/mssql/server:2019-latest'
+        docker exec -u 0 sql2019 bash -c 'chown mssql:root /data -R'
+        echo LOMBIQ_SHARED_TEST_CONFIGURATION="$PWD/.github/TestConfig.json" >> $GITHUB_ENV
+    - name: Set up Azurite via Docker
+      run: |
+        docker pull mcr.microsoft.com/azure-storage/azurite &&
+        docker run --name azurite -d -p 10000:10000 mcr.microsoft.com/azure-storage/azurite azurite-blob --blobHost 0.0.0.0 --blobPort 10000
+    - name: Unit Tests
```
Piedone: Since these are also UI tests:

```suggestion
    - name: Tests
```

.github/workflows/analyze_and_test.yml
```diff
@@ -0,0 +1,51 @@
...
+      run: |
+        (dotnet sln list |
+        grep '.Tests.' |
+        grep -v 'Lombiq.Tests.UI.csproj' |
+        grep -v 'Lombiq.Tests.csproj' |
+        sed -e 's/\\/\//g' -e '$atrue' -e 's/\(.*\)/echo "::group::\1" \&\&\ndotnet test -c Release --no-restore --no-build '"'\\1'"' \&\&\necho "::endgroup::" \&\&/' |
+        bash && (echo SUCCESS > .ui_test_result) || (echo FAIL > .ui_test_result)) | grep -v '^Connection refused \[::ffff:127.0.0.1\]';
+        grep SUCCESS .ui_test_result
```
Piedone: Make failure dumps available as an artifact: https://github.com/OrchardCMS/OrchardCore/pull/11194/files#diff-1153a5676594ad20522a7686e7e94a258891fe87611bafcd10ec0f3943c219b8

.github/workflows/analyze_and_test.yml
```diff
@@ -0,0 +1,51 @@
...
+        docker pull mcr.microsoft.com/azure-storage/azurite &&
+        docker run --name azurite -d -p 10000:10000 mcr.microsoft.com/azure-storage/azurite azurite-blob --blobHost 0.0.0.0 --blobPort 10000
+    - name: Unit Tests
+      run: |
+        (dotnet sln list |
+        grep '.Tests.' |
+        grep -v 'Lombiq.Tests.UI.csproj' |
+        grep -v 'Lombiq.Tests.csproj' |
+        sed -e 's/\\/\//g' -e '$atrue' -e 's/\(.*\)/echo "::group::\1" \&\&\ndotnet test -c Release --no-restore --no-build '"'\\1'"' \&\&\necho "::endgroup::" \&\&/' |
+        bash && (echo SUCCESS > .ui_test_result) || (echo FAIL > .ui_test_result)) | grep -v '^Connection refused \[::ffff:127.0.0.1\]';
```
Piedone: What on earth is happening here? :D Also, you can filter tests from within `dotnet test` too: https://docs.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests?pivots=mstest

Piedone (changes requested):

.github/workflows/analyze_and_test.yml
```diff
@@ -0,0 +1,56 @@
...
+        dotnet-version: 6.0.*
+    - name: Submodule Init
+      run: |
+        git submodule init && git submodule update
+    - name: Static Code Analysis
+      run: |
+        dotnet --list-sdks
+        npm install pnpm -g
+        until dotnet build Lombiq.OSOCE.sln --no-incremental -warnaserror /p:TreatWarningsAsErrors=true /p:RunAnalyzersDuringBuild=true -nologo -consoleLoggerParameters:NoSummary -verbosity:quiet >analysis.log 2>&1; do
+          grep 'error MSB3073: The command "npm run dotnet-postclean --if-present"' analysis.log >/dev/null 2>/dev/null && echo npm retry && continue
```
Piedone: We run a Gulp Extensions build first in TC. Might be needed.

.github/workflows/analyze_and_test.yml
sarahelsaig: Nope, because even this didn't fix it. Had to add a retry logic to make it reliable in the end. The affected projects already reference Lombiq.Npm.Targets so I have no idea why it's like this.

.github/workflows/analyze_and_test.yml
sarahelsaig: For some reason I thought it has to be in Debug. If it can be Release, then indeed there is no need for a second build.

It's quiet to eliminate irrelevant outputs. Really, all we care about is if there is some build error and if there isn't then there is no point outputting [this garbage information](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/runs/6430683041?check_suite_focus=true#step:6:12).

0liver: Good point about the `quiet` verbosity level. We might want to adapt that as the default in [our GitHub publish action](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/workflows/publish.yml#L12), as well.

.github/TestConfig.json
sarahelsaig: Could you explain why you think that would be nicer? I think adding some complicated variable like `$Lombiq_Tests_UI__SqlServerDatabaseConfiguration__ConnectionStringTemplate` would be less readable and it would be even worse if we had more settings, though just 2 is borderline.

.github/workflows/analyze_and_test.yml
sarahelsaig: That filter is not helpful because of the GitHub specific tricks that go on here. The goal is to place every test into a separate collapsible section within the step. See the generated script when you run the code up to this line in bash:
```sh
$ dotnet sln list |
>         grep '.Tests.' |
>         grep -v 'Lombiq.Tests.UI.csproj' |
>         grep -v 'Lombiq.Tests.csproj' |
>         sed -e 's/\\/\//g' -e '$atrue' -e 's/\(.*\)/echo "::group::\1" \&\&\ndotnet test -c Release --no-restore --no-build '"'\\1'"' \&\&\necho "::endgroup::" \&\&/'

echo "::group::src/Modules/Lombiq.TrainingDemo/Tests/Lombiq.TrainingDemo.Tests/Lombiq.TrainingDemo.Tests.csproj" &&
dotnet test -c Release --no-restore --no-build 'src/Modules/Lombiq.TrainingDemo/Tests/Lombiq.TrainingDemo.Tests/Lombiq.TrainingDemo.Tests.csproj' &&
echo "::endgroup::" &&
echo "::group::test/Lombiq.UITestingToolbox/Lombiq.Tests.UI.Shortcuts/Lombiq.Tests.UI.Shortcuts.csproj" &&
dotnet test -c Release --no-restore --no-build 'test/Lombiq.UITestingToolbox/Lombiq.Tests.UI.Shortcuts/Lombiq.Tests.UI.Shortcuts.csproj' &&
echo "::endgroup::" &&
echo "::group::test/Lombiq.UITestingToolbox/Lombiq.Tests.UI.AppExtensions/Lombiq.Tests.UI.AppExtensions.csproj" &&
dotnet test -c Release --no-restore --no-build 'test/Lombiq.UITestingToolbox/Lombiq.Tests.UI.AppExtensions/Lombiq.Tests.UI.AppExtensions.csproj' &&
echo "::endgroup::" &&
echo "::group::test/Lombiq.UITestingToolbox/Lombiq.Tests.UI.Samples/Lombiq.Tests.UI.Samples.csproj" &&
dotnet test -c Release --no-restore --no-build 'test/Lombiq.UITestingToolbox/Lombiq.Tests.UI.Samples/Lombiq.Tests.UI.Samples.csproj' &&
echo "::endgroup::" &&
echo "::group::src/Libraries/Lombiq.HelpfulLibraries/Lombiq.HelpfulLibraries.Tests/Lombiq.HelpfulLibraries.Tests.csproj" &&
dotnet test -c Release --no-restore --no-build 'src/Libraries/Lombiq.HelpfulLibraries/Lombiq.HelpfulLibraries.Tests/Lombiq.HelpfulLibraries.Tests.csproj' &&
echo "::endgroup::" &&
echo "::group::src/Modules/Lombiq.DataTables/Lombiq.DataTables/Tests/Lombiq.DataTables.Tests/Lombiq.DataTables.Tests.csproj" &&
dotnet test -c Release --no-restore --no-build 'src/Modules/Lombiq.DataTables/Lombiq.DataTables/Tests/Lombiq.DataTables.Tests/Lombiq.DataTables.Tests.csproj' &&
echo "::endgroup::" &&
echo "::group::test/Lombiq.OSOCE.Tests.UI/Lombiq.OSOCE.Tests.UI.csproj" &&
dotnet test -c Release --no-restore --no-build 'test/Lombiq.OSOCE.Tests.UI/Lombiq.OSOCE.Tests.UI.csproj' &&
echo "::endgroup::" &&
echo "::group::src/Modules/Lombiq.DataTables/Lombiq.DataTables/Tests/Lombiq.DataTables.Tests.UI/Lombiq.DataTables.Tests.UI.csproj" &&
dotnet test -c Release --no-restore --no-build 'src/Modules/Lombiq.DataTables/Lombiq.DataTables/Tests/Lombiq.DataTables.Tests.UI/Lombiq.DataTables.Tests.UI.csproj' &&
echo "::endgroup::" &&
echo "::group::src/Modules/Lombiq.VueJs/Lombiq.VueJs.Tests.UI/Lombiq.VueJs.Tests.UI.csproj" &&
dotnet test -c Release --no-restore --no-build 'src/Modules/Lombiq.VueJs/Lombiq.VueJs.Tests.UI/Lombiq.VueJs.Tests.UI.csproj' &&
echo "::endgroup::" &&
echo "::group::src/Themes/Lombiq.BaseTheme/Lombiq.BaseTheme.Tests.UI/Lombiq.BaseTheme.Tests.UI.csproj" &&
dotnet test -c Release --no-restore --no-build 'src/Themes/Lombiq.BaseTheme/Lombiq.BaseTheme.Tests.UI/Lombiq.BaseTheme.Tests.UI.csproj' &&
echo "::endgroup::" &&
echo "::group::src/Modules/Lombiq.ChartJs/Lombiq.ChartJs.Tests.UI/Lombiq.ChartJs.Tests.UI.csproj" &&
dotnet test -c Release --no-restore --no-build 'src/Modules/Lombiq.ChartJs/Lombiq.ChartJs.Tests.UI/Lombiq.ChartJs.Tests.UI.csproj' &&
echo "::endgroup::" &&
true
```

The `echo "::group::name` and echo "::endgroup::" are the magic strings that let you achieve this clean effect:
image

Also everything is joined with `&&` so if there is a test failure then the step is short cut to save on processing time.

.github/workflows/analyze_and_test.yml
sarahelsaig: ok I will try

.github/workflows/analyze_and_test.yml
0liver: This is quite magical.

.github/workflows/analyze_and_test.yml
sarahelsaig: life is kill 😭 
image

0liver: At least the error description is very verbose 😁

sarahelsaig: The last line makes me think it's so verbose because people were opening change request issues against this behavior.

.github/workflows/analyze_and_test.yml
sarahelsaig: Just needed to pre-build Gulp Extensions. Weird, becuase based on the project dependency trees that should already happen by that point.

.github/workflows/analyze_and_test.yml
Piedone: For Shouldly-using unit tests a Debug publish would be somewhat nicer but this doesn't matter much.

.github/TestConfig.json
Piedone: I learned about the central config file feature in the meantime. So to confirm, this file is used from all test projects and isn't copied around, right?

.github/workflows/analyze_and_test.yml
Piedone: I see, interesting. Please document this because just by looking at it I would have a very hard time figuring out what's the point here. Also, try to structure it more for readability.

.github/workflows/analyze_and_test.yml
Piedone: \> when were you when Dávid got upset?
\> i was sat at home, reviewing code, chilling to synthwave when e-mail with Dávid's comment
\> 'contains the following character: colon :'
\> 'no'

.github/TestConfig.json
sarahelsaig: Yes.

Piedone: Good then.

.github/workflows/analyze_and_test.yml
sarahelsaig: You can't safely put comments inside the script section (because it gets collapsed during parsing) so I had to write a wall of text above it :( . I have also formatted the code as much as possible, but sed and readability are not something you usually say in the same sentence.

Piedone: "sed"? :D

Yeah, I was meaning YAML comments anyway.

sarahelsaig: [sed](https://en.wikipedia.org/wiki/Sed) = stream editor, a classic Unix program. It does text replacing, splicing, etc and can be piped. Basically vim for shell scripting. You know you are writing sh when your `cat | sed` 😿

Piedone: Thanks, I hate it! :D

.github/workflows/analyze_and_test.yml
sarahelsaig: > chilling to synthwave

👍 👍 👍 👍

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,61 @@
...
+                -e '$atrue' \
+                -e 's/\(.*\)/echo "::group::\1" \&\&\ndotnet test -c Release --no-restore --no-build '"'\\1'"' \&\&\necho "::endgroup::" \&\&/' |
+            bash && (echo SUCCESS > .ui_test_result) || (echo FAIL > .ui_test_result)) |
+            grep -v '^Connection refused \[::ffff:127.0.0.1\]';
+        grep SUCCESS .ui_test_result
+    - uses: actions/upload-artifact@v2
+      if: failure()
+      with:
+          name: ui-test-failure
+          path: test/**/FailureDumps
```
Piedone: This also runs on unit test failure and causes a warning: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/2333914401 Maybe better run unit and UI tests in separate tests?

Piedone (changes requested):

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,61 @@
...
+      with:
+        dotnet-version: 6.0.*
+    - name: Submodule Init
+      run: |
+        git submodule init && git submodule update
+    - name: Build and Static Code Analysis
+      run: |
+        npm install pnpm -g
+        dotnet build src/Utilities/Lombiq.Gulp.Extensions/Lombiq.Gulp.Extensions.csproj --configuration Release
+        dotnet build Lombiq.OSOCE.sln --configuration Release -warnaserror /p:TreatWarningsAsErrors=true /p:RunAnalyzersDuringBuild=true -nologo -consoleLoggerParameters:NoSummary -verbosity:quiet
```
Piedone: We should also build the NuGet solution. That can perhaps be a second job?

.github/workflows/build_and_test.yml
sarahelsaig: That won't help. The `if: failure()` will evaluate true if any preceding tests failed. See this example:
https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/runs/6459542404?check_suite_focus=true

Piedone: Do you mean if any preceding step failed? This seems the be the case, [see docs](https://docs.github.com/en/actions/learn-github-actions/expressions#failure), but you can make steps depend on each other to make the UI testing and failure dump-creating step connected.

sarahelsaig: How do you tell UI and unit tests apart anyway?

sarahelsaig: I have found [this gem](https://docs.github.com/en/actions/learn-github-actions/expressions#hashfiles) and it seem to work. Doing `if: failure() && hashFiles('test/**/FailureDumps')` means upload-artifact only runs if there is something to dump.

Piedone: Great then! For the first question, I suppose based on name only.

sarahelsaig: > based on name only.

[shudder]

NuGetTest/.editorconfig
```diff
@@ -0,0 +1,287 @@
+# WARNING: Only edit this file in the Lombiq .NET Analyzers repository's folder. A copy of this file anywhere else will
```
Piedone: Why is this file needed? We shouldn't have to copy it. If Analyzers adds it then it should be gitignored.

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Tests/BehaviorModuleTests.cs
```diff
@@ -19,4 +21,17 @@ public BehaviorModuleTests(ITestOutputHelper testOutputHelper)
         ExecuteTestAfterSetupAsync(
             context => context.TestDataTableRecipeDataAsync(),
             browser);
+
+    [Theory, Chrome]
+    public Task BaseThemeShouldWork(Browser browser) =>
```
Piedone: Why did you move this from `BehaviorBaseThemeTests`? It was good there, it's not a module.

Piedone (changes requested):

Piedone:
You didn't address my feedback here.

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Tests/BehaviorModuleTests.cs
sarahelsaig: `BehaviorBaseThemeTests` is too specific. If it's not good here, then I'll move it to `BehaviorThemeTests` (without the "Base").

Piedone: Yeah, just a rename like that is good.

Piedone (approved):
Looks good but https://github.com/Lombiq/UI-Testing-Toolbox/pull/119#discussion_r877536815 affects this too.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,75 @@
...
+          -consoleLoggerParameters:NoSummary \
+          -verbosity:quiet
+    - name: Set up SQL Server via Docker
+      run: |
+        docker pull mcr.microsoft.com/mssql/server
+        docker run --name sql2019 -e 'ACCEPT_EULA=Y' -e 'SA_PASSWORD=Password1!' -p 1433:1433 -d 'mcr.microsoft.com/mssql/server:2019-latest'
+        docker exec -u 0 sql2019 bash -c 'mkdir /data; chmod 777 /data -R; chown mssql:root /data'
+    - name: Set up Azurite via Docker
+      run: |
+        docker pull mcr.microsoft.com/azure-storage/azurite &&
```
0liver, line 44: Why do you chain with `&&` here but not in the SQL server setup [above](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/41/files#diff-48c0ee97c53013d18d6bbae44648f7fab9af2e0bf5b0dc1ca761e18ec5c478f2R39)?

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,75 @@
...
+        docker run --name azurite -d -p 10000:10000 mcr.microsoft.com/azure-storage/azurite azurite-blob --blobHost 0.0.0.0 --blobPort 10000
+    - name: Tests
+      # This step gets the list of all projects, selects the test projects (excluding the two test libraries) and then
+      # uses sed to create an inline bash script. What this script does:
+      # - Uses 'dotnet test -c Release --no-restore --no-build' to start up each project.
+      # - Before the test outputs '::group::project name' and after outputs '::endgroup::'. This tells GitHub that the
+      #   outputs should be placed into a collapsible region on the Actions web UI.
```
0liver: I needed some clarification on this comment, so here's what I came up with:
```suggestion
      # - The test output is wrapped by '::group::<project name>' before and '::endgroup::' after it. This tells GitHub
      #   to place each test's output into a collapsible region on the Actions web UI.
```

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,75 @@
...
+        (dotnet sln list |
+            grep '.Tests.' |
+            grep -v 'Lombiq.Tests.UI.csproj' |
+            grep -v 'Lombiq.Tests.csproj' |
+            sed -e 's/\\/\//g' \
+                -e '$atrue' \
+                -e 's/\(.*\)/echo "::group::\1" \&\&\ndotnet test -c Release --no-restore --no-build '"'\\1'"' \&\&\necho "::endgroup::" \&\&/' |
+            bash && (echo SUCCESS > .ui_test_result) || (echo FAIL > .ui_test_result)) |
+            grep -v '^Connection refused \[::ffff:127.0.0.1\]';
+        grep SUCCESS .ui_test_result
```
0liver, line 70: I'm seriously impressed.

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/.htmlvalidate.json
```diff
@@ -14,5 +14,16 @@
...
     "wcag/h71": "off"
-  }
+  },
+
+  "elements": [
+    "html5",
+    {
+      "form": {
+        "attributes": {
+          "method": [ "get", "post", "GET", "POST" ]
```
0liver: Is this now supposed to be unnecessary with the [latest version](https://gitlab.com/html-validate/html-validate/-/releases/v7.1.1)?

0liver (changes requested):

.github/workflows/build_and_test.yml
sarahelsaig: By accident.

.github/workflows/build_and_test.yml
sarahelsaig: Hmm, some nuance is lost in translation, but I guess it carries the gist better so I'm in favor of it.

Lombiq.OSOCE.sln
```diff
@@ -122,6 +124,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Themes", "Themes", "{93A678
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Lombiq.BaseTheme.Tests.UI", "src\Themes\Lombiq.BaseTheme\Lombiq.BaseTheme.Tests.UI\Lombiq.BaseTheme.Tests.UI.csproj", "{2BF13E3A-B54D-4352-812E-9A867A801FFE}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Lombiq.HelpfulLibraries.Cli", "src\Libraries\Lombiq.HelpfulLibraries\Lombiq.HelpfulLibraries.Cli\Lombiq.HelpfulLibraries.Cli.csproj", "{F4B0F77A-4EF0-40A9-B1C0-9294B397BA36}"
```
Piedone, line 127: Add a corresponding line to the Readme under "Included Projects".

0liver (approved):

0liver (approved):

PR #62, opened 2022-03-02
INFRA-97: Running UI Tests in a container

BenedekFarkas:
INFRA-97


BenedekFarkas:
We don't need this for now.

PR #75, opened 2022-03-22
OSOE-83: Publish NodeJs.Extensions to NuGet

0liver:
OSOE-83


Lombiq.OSOCE.sln
```diff
@@ -69,20 +67,28 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Lombiq.Privacy", "src\Modul
...
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Lombiq.HelpfulLibraries.OrchardCore", "src\Libraries\Lombiq.HelpfulLibraries\Lombiq.HelpfulLibraries.OrchardCore\Lombiq.HelpfulLibraries.OrchardCore.csproj", "{CA011075-6748-4EB4-9C72-C8FD7319FEFD}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Lombiq.HelpfulLibraries.OrchardCore.Testing", "src\Libraries\Lombiq.HelpfulLibraries\Lombiq.HelpfulLibraries.OrchardCore.Testing\Lombiq.HelpfulLibraries.OrchardCore.Testing.csproj", "{95D73C6A-B81E-44A3-9376-9CFEFDE3B314}"
+EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Lombiq.NodeJs.Extensions.Samples", "src\Utilities\Lombiq.NodeJs.Extensions\Lombiq.NodeJs.Extensions.Samples\Lombiq.NodeJs.Extensions.Samples.csproj", "{D4D7F555-3AA2-43C8-8AD2-64136A6C8053}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Lombiq.NodeJs.Extensions.Samples.NuGet", "src\Utilities\Lombiq.NodeJs.Extensions\Lombiq.NodeJs.Extensions.Samples.NuGet\Lombiq.NodeJs.Extensions.Samples.NuGet.csproj", "{94C52D1B-7BA3-4559-B180-BE10A12BFA5D}"
```
Piedone: This needs to rather be referenced from the NuGet solution.

Piedone (changes requested):

Lombiq.OSOCE.sln
0liver: Where should I put it?
image

It's not really a module.

Piedone: Match the structure in the root solution.

0liver: Do you expect anything else to happen here with that sample module?

Also, I'm a bit confounded by the fact that an error in the SCSS throws a build error when I build the solution, but when I hit F5 it still runs. Is that normal?

Piedone: What do you mean by anything else? The NuGet solution is to test if our projects still work when in NuGet form, that's the goal here too.

Perhaps it runs with the latest successful build actually? Such build errors shouldn't be different than those produced by the C# compiler.

0liver: Well, the `Samples.NuGet` project doesn't do much else than reference the NE NuGet package and build its assets during the build. If that's enough, good.

For the build: I guess, I'll have to add a project reference to `Samples.NuGet` in `NuGet.Web` 😉

Piedone: Yes. This is just about building assets.

0liver: I've got the run failure in case of npm build errors covered, now.

Piedone (approved):

Piedone (approved):

Piedone (approved):

Piedone (approved):

Directory.Build.props
```diff
@@ -1,16 +1,16 @@
...
-  
+
   <PropertyGroup>
     <!-- This setting prevents .stylelintrc files from being created in projects that use Lombiq Node.js Extensions.
          Instead, we use a global .stylelintrc file at the solution root. -->
     <NodeJsExtensionsCreateStylelintConfigurationFile>false</NodeJsExtensionsCreateStylelintConfigurationFile>
-  </PropertyGroup> 
-  
+  </PropertyGroup>
+
```
0liver, line 9: Oops!

Piedone: No worries.

PR #98, opened 2022-04-26
OCC-3: Update Lombiq.Lombiq.Analyzers

sarahelsaig:
[OCC-3](https://lombiq.atlassian.net/browse/OCC-3)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,4 +1,4 @@
-<Project>  
+<Project>
   <!-- Copying the Lombiq logo as an icon to all NuGet-published projects. -->
   <Target Name="Copy NuGet icon" AfterTargets="AfterResolveReferences" Condition="'$(PackageLicenseFile)' != ''">
     <Copy SourceFiles="$(MSBuildThisFileDirectory)LombiqSymbolLogoColorWhiteFlaskSquareForNuGetPackages.png" DestinationFiles="$(ProjectDir)NuGetIcon.png" SkipUnchangedFiles="true" />
@@ -9,6 +9,6 @@
   <!--<PropertyGroup>
     <NuGetBuild>true</NuGetBuild>
   </PropertyGroup>-->
-  
-  <Import Project="tools/Lombiq.Analyzers/Build.props" Condition="'$(NuGetBuild)' == 'true'" />
+
+  <Import Project="tools/Lombiq.Analyzers/Build.props" Condition="'$(NuGetBuild)' != 'true'" />
 </Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,5 +1,5 @@
 <Project>
-  <Import Project="tools/Lombiq.Analyzers/Build.props"/>
+  <Import Project="tools/Lombiq.Analyzers/Build.props" Condition="'$(NuGetBuild)' == 'true'" />
   
   <!-- Copying the Lombiq logo as an icon to all NuGet-published projects. -->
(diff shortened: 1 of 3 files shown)


Directory.Build.props
```diff
@@ -1,5 +1,5 @@
 <Project>
-  <Import Project="tools/Lombiq.Analyzers/Build.props"/>
+  <Import Project="tools/Lombiq.Analyzers/Build.props" Condition="'$(NuGetBuild)' == 'true'" />
```
0liver: To me, the commit message reads as if the condition should be negated:
> Don't include submodule analyzer in NuGet-specific solution. 

What am I not getting?

Directory.Build.props
```diff
@@ -11,4 +9,6 @@
   <!--<PropertyGroup>
     <NuGetBuild>true</NuGetBuild>
   </PropertyGroup>-->
+  
+  <Import Project="tools/Lombiq.Analyzers/Build.props" Condition="'$(NuGetBuild)' == 'true'" />
```
0liver: Looks like the condition should be removed again, right?

image

0liver (changes requested):

Directory.Build.props
sarahelsaig: No, it's my mistake. I've fixed it and verified that the analyzers work as expected both solutions.
(here "expected" means "correctly" with submodules, and "broken" with NuGet, [as explained here](https://teams.microsoft.com/l/message/19:525a1a3656c845ae92c5e2c0d3617cc7@thread.skype/1651237342049?tenantId=959ede5a-38e9-44e6-adc8-85553beb8220&groupId=d6a5d775-6fb9-41dc-a7bb-481d4f884046&parentMessageId=1651228778188&teamName=Lombiq%20Developers&channelName=General&createdTime=1651237342049))

Directory.Build.props
sarahelsaig: The condition should stay, becuase it is correct (now that I have fixed it). The analyzer itself is broken on NuGet, [but that's a separate issue](https://lombiq.atlassian.net/browse/OSOE-110).

NuGetTest/src/Modules/Lombiq.OSOCE.NuGet.TestModule/Controllers/AnalyzersTestController.cs
```diff
@@ -8,9 +8,9 @@ public class AnalyzersTestController : Controller
 {
     private readonly IClock _clock;
 
-    // Since the Analyzers NuGet package is added, you can see that it works like this: Disable the IDE0021 rule in the
-    // root Analyzers OrchardCore.ruleset file, then change this constructor to use a block body. You should get an
-    // analyzer violation since the expression-bodied rule is still enforced from the package
+    // Since the Analyzers NuGet package is added, you can see that it works like this: change this constructor to use a
+    // block body. You should get an analyzer violation since the expression-bodied rule is still enforced from the
+    // package
```
0liver:
```suggestion
    // Since the Analyzers NuGet package is added, you can see that it works like this: Change this constructor to use a
    // block body. You should get an analyzer violation since the expression-bodied rule is still enforced from the
    // package.
```

Mr. Z insists on Capitalization after the colon, so please adhere. I insist on a dot at the end of the sentence.

0liver (changes requested):

0liver (approved):

PR #109, opened 2022-05-16
OSOE-110: ⭐ Fix that the NuGet version of Lombiq Analyzer doesn't detect any analyzer violations

sarahelsaig:
OSOE-110


sarahelsaig:
At this point the test _should_ fail. Specifically I expect that one of the two cases of `AnalyzerViolationShouldBeReported` passes and the other fails, since the current state is that submodule works but NuGet does not.

This actually happens on the GitHub workflow, but the TC doesn't include the new _Lombiq.Analyzers.Tests_ project at all, so it passes. Any idea what I'm missing?

Piedone:
It only looks under the src folder for test projects. I've changed the config to look under _tools_ too, and restarted the build. Now it should fail.

sarahelsaig:
Thanks, now everything is as it should be.
image

Piedone:
I've reset the config to the original now.

NuGetTest/src/Themes/Lombiq.OSOCE.NuGet.TestTheme/Lombiq.OSOCE.NuGet.TestTheme.csproj
```diff
@@ -15,10 +15,6 @@
     <EmbeddedResource Remove="package.json" />
   </ItemGroup>
 
-  <ItemGroup>
-    <Content Include="package.json" />
```
sarahelsaig, line 19: This was somehow a duplicate. Not sure which one should be kept or if it matters.

Directory.Build.props
```diff
@@ -10,5 +14,6 @@
     <NuGetBuild>true</NuGetBuild>
   </PropertyGroup> -->
 
-  <Import Project="tools/Lombiq.Analyzers/Build.props" Condition="'$(NuGetTestbed)' != 'true'" />
+  <Import Project="tools/Lombiq.Analyzers/Lombiq.Analyzers/Build.props"
+          Condition="'$(NuGetTestbed)' != 'true' AND '$(SkipRootBuildProps)' != 'true'"/>
```
0liver: Both of these conditions should turn out unnecessary, since it's the existence of a _Directory.Build.props_ further down the folder hierarchy that stops importing those further up ([docs](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022#search-scope)).

NuGetTest/src/Themes/Lombiq.OSOCE.NuGet.TestTheme/Lombiq.OSOCE.NuGet.TestTheme.csproj
0liver: I don't know either right now.

0liver (changes requested):

Directory.Build.props
sarahelsaig: It's also used for something infrastructure-related. ([see issue description](https://lombiq.atlassian.net/browse/OSOE-110#:~:text=ask%20the%20infrastructure%20team%20to%20replace%20-p%3Anugettestbed%3Dtrue%20with%20-p%3Askiprootbuildprops%3Dtrue%20in%20the%20gulp%20extension%20build%20of%20the%20build%20and%20test%20(developer)%20configuration.%20see%20this%20teams%20thread.))

0liver (approved):

PR #110, opened 2022-05-17
CHOIR-4: Making logging actually work

porgabi:
CHOIR-4

### DIFF
--- a/src/Lombiq.OSOCE.Web/HostBuilderExtensions.cs
+++ b/src/Lombiq.OSOCE.Web/HostBuilderExtensions.cs
@@ -18,7 +18,7 @@ public static class HostBuilderExtensions
         {
             var environment = context.HostingEnvironment;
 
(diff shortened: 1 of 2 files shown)


src/Lombiq.OSOCE.Web/HostBuilderExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    public static IHostBuilder UseNLogHost(this IHostBuilder builder)
+    {
+        LayoutRenderer.Register<TenantLayoutRenderer>(TenantLayoutRenderer.LayoutRendererName);
+        builder.UseNLog();
+        builder.ConfigureAppConfiguration((context, _) =>
+        {
+            var environment = context.HostingEnvironment;
+
+            environment.ConfigureNLog($"{environment.ContentRootPath}{Path.DirectorySeparatorChar}NLog.config");
```
0liver: I think, `Path.Combine()` is for that.

PR #115, opened 2022-05-24
OSOE-111: Detect visual errors (visual verification testing) in the UI Testing Toolbox

dministro:
OSOE-111


.github/workflows/build-and-test.yml
```diff
@@ -124,6 +124,7 @@ jobs:
             export Lombiq_Tests_UI__DockerConfiguration__ContainerName="sql2019"
         else
             export Lombiq_Tests_UI__SqlServerDatabaseConfiguration__ConnectionStringTemplate="Server=.\\SQLEXPRESS;Database=LombiqUITestingToolbox_{{id}};Integrated Security=True;MultipleActiveResultSets=True;Connection Timeout=60;ConnectRetryCount=15;ConnectRetryInterval=5"
+            powershell Set-DisplayResolution -Width 1920 -Height 1080 -Force
```
Piedone: This shouldn't be necessary as UI testing shouldn't rely on the screen resolution. That's why we also have `SetBrowserSize()`.

Piedone (changes requested):
Merge from dev, please.

.github/workflows/build-and-test.yml
dministro: Unfortunately, this or something similar is necessary, because, without this, the browser window size is smaller than expected on windows-latest.
Below you can find screenshots and sizes before and after setting screen resolution on windows-latest.

**ubuntu-latest:**
size: 1600 ×900
ubuntu-latest_navbar_full_screen

**windows-latest**-before setting screen resolution
size: 1028 × 655
windows-latest_navbar_full_screen

**windows-latest**-after setting screen resolution
size: 1600 ×900
windows-latest_after_set_screen_size_navbar_full_screen

Piedone: Please look into how this works with GitHub Actions, and why this difference between platforms is there.

dministro: **Maximum window size on windows-latest:**
If I remember well, the maximum window size can't be larger than the desktop size(plus some pixels) on windows.
I didn't find official documentation about this, but I tested it in non-headless mode:
My desktop size: 3840×1200
Requested window size: context.SetBrowserSize(CommonDisplayResolutions.FiveK)
The resulted window size: 3860×1220
image

**GH windows-latest default screen resolution:**
I didn't find official documentation about this, but here is an issue(related to Windows 2019 and screen size): https://github.com/actions/virtual-environments/issues/2935

I'll give it a try to run UI tests in headless mode on windows-latest. As I see in the workflow and configuration they are started in non-headless mode.

dministro: It's not surprising, but all tests passed on windows-latest in headless mode without setting screen resolution.

So, which mode is preferred on windows-latest? headless or non-headless?

Piedone: Great then! Headless all the way. However, how was it not headless before? Hmm, `UITestBase` makes it default to `false` in both test projects (due to the global default being `true`). I think we did that to aid in inspecting them locally, due to the demonstration purpose the two test projects serve. So, you can just set that config as `true` in the GitHub build's `Tests` step. (In TeamCity it's set as such from its config.)

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Lombiq.OSOCE.NuGet.Tests.UI.csproj
```diff
@@ -11,16 +11,16 @@
 
   <ItemGroup>
     <Content Include=".htmlvalidate.json">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
Piedone, line 14: Why is this necessary?

NuGetTest/src/Themes/Lombiq.OSOCE.NuGet.TestTheme/Lombiq.OSOCE.NuGet.TestTheme.csproj
```diff
@@ -25,10 +25,10 @@
 
   <ItemGroup>
     <PackageReference Include="Lombiq.Npm.Targets" Version="1.0.1" />
-    <PackageReference Include="OrchardCore.Theme.Targets" Version="1.3.0" />
-    <PackageReference Include="OrchardCore.ContentManagement" Version="1.3.0" />
-    <PackageReference Include="OrchardCore.DisplayManagement" Version="1.3.0" />
-    <PackageReference Include="OrchardCore.ResourceManagement" Version="1.3.0" />
+    <PackageReference Include="OrchardCore.Theme.Targets" Version="1.4.0" />
```
Piedone: Thanks for noticing these!

Piedone (changes requested):

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Lombiq.OSOCE.NuGet.Tests.UI.csproj
dministro: This was an accident.

This change was because there are 3 .htmlvalidate.json under OSOCE and all of it is set to `PreserveNewest`. One of them is in Lombiq.Tests.UI NuGet package, is a dependency of Lombiq.OSOCE.NuGet.Tests.UI project, which contains .htmlvalidate.json too. After I created a new prerelease from Lombiq.Tests.UI project, it become newer, and this caused problems on the local run of NuGet tests.

Piedone:
Merge from dev, please.

Piedone (approved):

.github/workflows/build-and-test.yml
```diff
@@ -141,6 +141,11 @@ jobs:
                 Value = 'sql2019'
             }
         }
+        
+        Configure-Environment @{
+            Path = 'Env:\Lombiq_Tests_UI__BrowserConfiguration__Headless'
+            Value = 'true'
+        }
```
Piedone: You'll need to add this to https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/test-dotnet/run-tests.ps1 after merging from dev. And there it would be simpler to just write `$Env:Lombiq_Tests_UI__BrowserConfiguration__Headless = "true"` instead, and changing the existing `Configure-Environment`s to that too.

dministro: Thank you for the info.

Piedone (approved):

Piedone (approved):

Piedone (changes requested):
Merge from dev, please.

PR #124, opened 2022-05-31
OSOE-125: ⭐ Add JavaScript processing functionality to Node Extensions

0liver:
OSOE-125

- Successful UIKit publish: https://github.com/Lombiq/Orchard-UIKit/actions/runs/2560234955
- Successful NE publish: https://github.com/Lombiq/NodeJs-Extensions/actions/runs/2560256111


Piedone (changes requested):
The NuGet build shouldn't be affected by this work and thus shouldn't fail (previously under OSOE-92 it didn't fail either). In TC it fails with a build error, and in GitHub Actions with a test (note that it's the Ubuntu build, the Windows one most possibly would fail in the same way as in TC, you can temporarily set `fail-fast: false` to get all builds to run and thus see it, even if one of them fail, see [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast)).

Under [the diff of this branch to dev](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/compare/dev...issue/OSOE-125) you can see the affected NuGet solution changes; apart from adding `Lombiq.NodeJs.Extensions.Samples.NuGet` to the solution, I don't see why any of these are necessary (only talking about the NuGet solution changes, not the root solution).

.github/workflows/build-and-test.yml
```diff
@@ -37,7 +37,15 @@ jobs:
     - name: Build and Static Code Analysis
       working-directory: ${{ matrix.directory }}
       run: |
-        npm install pnpm -g
+        # Map current working directory to drive letter to shorten the absolute path of some deeply nested files.
```
Piedone: Note that note yet, but when merging `task/node-extensions` to `dev`, this will need to be added in a DRY was to https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/workflows/build-and-test-orchard-core.yml (and other changes in this workflow too).

pnpm-lock.yaml
```diff
@@ -0,0 +1,1144 @@
+lockfileVersion: 5.3
```
Piedone, line 1: I get these diffs here after building a cleaned clone (`lockfileVersion` also changes in UI Testing Toolbox):

image

.eslintrc.json
```diff
@@ -0,0 +1,11 @@
+{
```
Piedone, line 1: Why does this needs to be .json?

Piedone (changes requested):

pnpm-lock.yaml
0liver: This happens when **pnpm@v7** is used. I've refrained from updating to v7 for now, because it brought some breaking changes to how scripts are called, which I'd rather address in a separate issue.

.eslintrc.json
0liver: It does not need to be .json - I just chose that to aid JSON syntax highlighting in editors. Objections?

pnpm-lock.yaml
Piedone: Which version should we use, then? Because v7 has been the recommendation under https://lombiq.atlassian.net/wiki/spaces/DEV/pages/1748369460/Setting+up+Node.js+with+Visual+Studio for a while.

.eslintrc.json
Piedone: OK then, but then can't we do the same for stylelintrc?

0liver: Yes, we will. I will refactor the styles pipeline as part of OSOE-84 to reflect the improvements I was able to make during my work on the JS pipeline. All differences will be addressed there.

pnpm-lock.yaml
0liver: Well, pnpm 7.0.0 has only been released [2 months ago](https://github.com/pnpm/pnpm/releases/tag/v7.0.0), so I would not expect us to jump on it right away. Our docs don't mention a version number for **pnpm**, but I can add one for now, until an upgrade from v6 to v7 has been successfully implemented for Node.js Extensions.

Piedone: Yeah, it was the npm, not pnpm version. However, then please add a pnpm version to the wiki.

0liver: Done.

Piedone: Shouldn't it be a specific version, just to be sure?

0liver: The latest v6 will be fine. There shouldn't be any breaking changes, and I don't mean the semantics that the .NET SDK has introduces for patch version updates 😁

Piedone (approved):

.github/workflows/build-and-test.yml
```diff
@@ -8,17 +8,16 @@ on:
       - dev
 
 jobs:
-  call-build-and-test-workflow:
+  call-build-workflow:
     name: Build and Test - root solution
-    uses: Lombiq/GitHub-Actions/.github/workflows/build-and-test-orchard-core.yml@dev
+    uses: Lombiq/Open-Source-Orchard-Core-Extensions/.github/workflows/build-dotnet-with-pnpm.yml@issue/OSOE-125
```
Piedone, line 13: Keep in mind to change these once you start OSOE-84 again. BTW I think for references within the same caller repo you don't need such branch references, you can just reference the file like we do in the GHA repo (those only need branch references because they're called from other repos).

PR #129, opened 2022-06-03
OSOE-127: Upgrade OSOCE to Orchard Core 1.4

### LINKED ISSUE #125: Upgrade to Orchard Core 1.4
[Orchard Core 1.4](https://github.com/OrchardCMS/OrchardCore/releases/tag/v1.4.0) is now out. Let's upgrade to it. See [release notes](https://docs.orchardcore.net/en/latest/docs/releases/1.4.0/).

DemeSzabolcs:
Resolves #125 
OSOE-127


src/Lombiq.OSOCE.Web/HostBuilderExtensions.cs
```diff
@@ -1,41 +0,0 @@
-using Microsoft.Extensions.Hosting;
```
DemeSzabolcs, line 1: This (HostBuilderExtensions.cs) is now in OC: https://github.com/OrchardCMS/OrchardCore/pull/11604/files

src/Lombiq.OSOCE.Web/Recipes/Lombiq.OSOCE.Tests.recipe.json
```diff
@@ -85,6 +85,114 @@
         ]
       }
     },
+    {
```
sarahelsaig, line 89: Could you include a comment on why this is necessary?

sarahelsaig (changes requested):

src/Lombiq.OSOCE.Web/Recipes/Lombiq.OSOCE.Tests.recipe.json
DemeSzabolcs: Sure. So for some reason, after updating the packages, these roles (that I added) were not created during testing by default thus, some tests failed because the homepage was not displaying (instead of it, it displayed a login page, because an anonymous user couldn't view the homepage), so I needed to add them. This same thing already happened to me in another project a while ago.

Should I add a comment about this to the code too?

sarahelsaig: Is this a problem during normal setup too or just during testing? 

> Should I add a comment about this to the code too?

Yes, please.

DemeSzabolcs: > Is this a problem during normal setup too or just during testing?

It's only present during testing.

PR #131, opened 2022-06-07
OSOE-126: Present a nicer test report for GitHub Actions builds

Luko6:
OSOE-126


.github/workflows/test-report.yml
```diff
@@ -0,0 +1,20 @@
+﻿name: Report
+
+on:
+  workflow_run:
+    workflows: ['Build and Test']
```
sarahelsaig:
```suggestion
    workflows: [Build and Test]
```

[Based on the samples](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) you don't need the apostrophes. Could it be that this workflow is now waiting for a workflow called "'Build and Test'"?

Luko6: It's waiting for the workflow already present yes.

Luko6: It's strating after the whole workflow is done, so after all 4 have either failed or succeeded.

.github/workflows/test-report.yml
```diff
@@ -0,0 +1,20 @@
...
+    workflows: [Build and Test]
+    types:
+      - completed
+    
+jobs:
+  report:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: phoenix-actions/test-reporting@v8
```
Piedone: You don't need to put this into a separate workflow. This should rather be a step in build-and-test.

Luko6: This documentation should answer why it's seperate:

> Workflows triggered by pull requests from forked repositories are executed with read-only token and therefore can't create check runs. To workaround this security restriction, it's required to use two separate workflows
https://github.com/marketplace/actions/test-reporting#recommended-setup-for-public-repositories

Piedone: Then add this as a comment there as well.

Piedone: On a second thought we should rather keep this in the original workflow still. We'll use this in all GitHub Actions solution builds of ours, and it'll be an issue to maintain multiple workflow files instead of one. "Workflows triggered by pull requests from forked repositories are executed with read-only token and therefore can't create check runs." is not really true like this, since workflows do run from forks under PRs to the original repository, you just need to approve it.

Luko6: Allright.

.github/workflows/build-and-test.yml
```diff
@@ -150,3 +150,11 @@ jobs:
         name: ui-test-failure-dump-${{ matrix.directory }}-${{ matrix.os }}
         path: ${{ matrix.directory }}/test/**/FailureDumps/
         if-no-files-found: ignore
+
+    # Test Report
```
Piedone: This is redundant with the step name.

.github/workflows/build-and-test.yml
```diff
@@ -150,3 +150,11 @@ jobs:
...
+    - name: Test Report
```
Piedone, line 199: Check out whether this runs when tests fail. [See the `if`](https://github.com/marketplace/actions/test-reporting#example).

Piedone, line 199: Add docs about what these Test Results builds are under https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2455732262/Continuous+Integration#GitHub-Actions-overview.

Piedone, line 199: Can we do something about the horizontal scrolling in the tables here, even if there's enough space? 
image

Piedone (changes requested):

Piedone:
You need to merge from dev.

.github/workflows/build-and-test.yml
Luko6: Shouldn't this be a part of the "After Resolve" section? I'll write the text for it though

Luko6: Not really, the container has a max width of 768px (coming from the `container-md` class), and there is no setting to modify this.

Piedone: Yeah, good point.

PR #135, opened 2022-06-13
OSOE-109: Consolidate casing of constants and @injected dependencies in Views

VTamas1:
OSOE-109


Piedone (changes requested):
Merge from dev, please.

PR #149, opened 2022-07-05
OSOE-150: ⭐ Add PowerShell static code analysis to OSOCE

sarahelsaig:
OSOE-150
- https://github.com/Lombiq/GitHub-Actions/pull/23
- https://github.com/Lombiq/PowerShell-Analyzers/pull/1
- https://github.com/Lombiq/Helpful-Libraries/pull/149
- https://github.com/Lombiq/NPM-Targets/pull/20

[Example of it working.](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/runs/7197098688?check_suite_focus=true) Also see the unit tests.


.github/workflows/build-and-test.yml
```diff
@@ -10,15 +10,15 @@ on:
 jobs:
   call-build-and-test-workflow:
     name: Build and Test - root solution
-    uses: Lombiq/GitHub-Actions/.github/workflows/build-and-test-orchard-core.yml@dev
+    uses: Lombiq/GitHub-Actions/.github/workflows/build-and-test-orchard-core.yml@issue/OSOE-150
```
Piedone: Demonstrate an analyzer violation during the GHA build, please.

src/Lombiq.OSOCE.Web/Lombiq.OSOCE.Web.csproj
```diff
@@ -6,6 +6,8 @@
     <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
   </PropertyGroup>
 
+  <Import Project="..\..\tools\Lombiq.Analyzers.PowerShell\Lombiq.Analyzers.PowerShell\Lombiq.Analyzers.PowerShell.targets" Condition="'$(GITHUB_ENV)' == ''" />
```
Piedone: Why do we need this, and why the distinction with `GITHUB_ENV`?

Piedone (changes requested):

.github/workflows/build-and-test.yml
sarahelsaig: I have already linked it in the issue description.

src/Lombiq.OSOCE.Web/Lombiq.OSOCE.Web.csproj
sarahelsaig: When you run the build in GitHub Actions, it should use the workflow command output format, whereas this uses the MSBuild output format. Unfortunately GH doesn't automatically convert those into error annotations.

.github/workflows/build-and-test.yml
Piedone: In the PR description, but I see, thanks.

src/Lombiq.OSOCE.Web/Lombiq.OSOCE.Web.csproj
Piedone: Do we actually need a separate PS analyzer step in the GH workflow if we have MSBuild integration?

sarahelsaig: If you suppress the logs like we do in the GH workflow (with `-verbosity:quiet`) then also suppresses any stdout and stderr outputs from the script called by `<Exec>`. This means we can only emit MSBuild errors, because MSBuild treats those differently, but we can't emit GH workflow commands such as `::error::My message.`. When I try to use the MSBuild commands in the GH actions, they display on the console output but aren't interpreted as errors so not great.

Piedone: Shouldn't we emit MSBuild errors, then? Since this targets file integrates PS analysis with the build, it could also integrate with the output there, like ESLint and Stylelint. I.e. use warnings by default, which will be turned into errors when we build in CI.

sarahelsaig: Yeah, I was also considering that. I will look into a way to turn MSBuild errors into GH errors in `dotnet-build`. Then we can stick with just one error type per tooling.

sarahelsaig: Works:
image
image

Piedone: Looks awesome!

.github/workflows/build-and-test.yml
Piedone: I get these errors locally but the GHA build passes:

image

BTW shouldn't these be warnings locally instead?

sarahelsaig: No, becuase if you return warnings for the items, but still emit a nonzero exit code that's pointless. But if the analyzer returns zero then it won't emit the warnings a second time on incremental build (because the timestamp would be updated) which would be confusing.

Piedone: These errors are still there. And again, shouldn't this cause the GHA build to fail too?

sarahelsaig: I'm not getting any of these locally. I'm pretty sure these hint to version differences between Windows PowerShell and PowerShell 7. The "in this version" must mean 5.x, right? Also I remember, that `curl` was an alias in Windows PowerShell (which was dumb, they shouldn't have done that), but it's not an alias in PS7. MSBuild prefers `pwsh` when it's available (so always on GHA), that's why the builds don't fail.

Both of the affected files are in the _Lombiq.GitHub.Actions_ repo, where all scripts are only invoked with `pwsh`. (and `curl` is the actual cURL program) I don't think we need to worry about how those scripts behave in older versions since we won't ever use those.

(this doesn't apply to Utility-Scripts, where scripts has to support Windows PowerShell, but there the analysis runs on both versions of PS, so I think we have good coverage)

Piedone: Hmm, it should have been a version issue. I installed the latest PS 7 and it works now.

Piedone (approved):

Piedone (approved):

Piedone (approved):

Piedone (approved):

PR #153, opened 2022-07-12
OSOE-125: Add path shortening during GitHub Actions build

0liver:
OSOE-125

https://github.com/Lombiq/GitHub-Actions/pull/17


Piedone (changes requested):
This doesn't reference the issue branch in the workflow, thus it doesn't use https://github.com/Lombiq/GitHub-Actions/pull/17.

Piedone (approved):

PR #165, opened 2022-07-30
OCC-11: Updating documentation for JS scripts

DemeSzabolcs:
OCC-11


sarahelsaig:
TeamCity is expected to fail because it still expects Gulp Extensions. The rest is ok!

PR #194, opened 2022-08-31
NEST-316: Re-implement Azure Deploy build with GitHub Actions

DemeSzabolcs:
NEST-316


Piedone (approved):

Piedone (approved):

Piedone (changes requested):
Point the build workflows to your branch in GitHub Actions, so the build runs your code. Same for NEST.

Also keep the submodule up to date here.

Piedone (changes requested):
As mentioned, point the build (and test) workflows here to your branch in the GitHub Actions repo, so the build runs your code. Same for NEST. Currently, these don't run the build workflow that you modified and thus we're not testing it.

Piedone (approved):

PR #217, opened 2022-09-29
NEST-282: FeaturesGuard Additions

porgabi:
NEST-282


src/Lombiq.OSOCE.Web/appsettings.json
```diff
@@ -22,6 +22,14 @@
...
       "EnableLoggingTestMiddleware": true
     },
+    "OrchardCore_DataProtection_Azure": {
+      "ConnectionString": "UseDevelopmentStorage=true",
+      "ContainerName": "osocedataprotection"
+    },
+    "OrchardCore_Media_Azure": {
+      "ConnectionString": "UseDevelopmentStorage=true",
+      "ContainerName": "osocemedia"
+    },
```
wAsnk, line 32: You need this only in development

test/Lombiq.OSOCE.Tests.UI/Helpers/AssertAppLogsHelpers.cs
```diff
@@ -0,0 +1,63 @@
...
+                !message.Contains("System.IO") &&
+                !message.Contains("|WARN|") &&
+                !message.Contains("|Lombiq.TrainingDemo.Services.DemoBackgroundTask|ERROR|Expected non-error") &&
+                !message.Contains("Azure Media Storage is enabled but not active because the 'ContainerName' is missing or empty") &&
+                !message.Contains("Azure Media Storage is enabled but not active because the 'ConnectionString' is missing") &&
+                !message.Contains("OrchardCore.Media.Core.DefaultMediaFileStoreCacheFileProvider.TryDeleteDirectoryAsync") &&
+                !message.Contains("OrchardCore.Media.Core.DefaultMediaFileStoreCacheFileProvider|ERROR|Error deleting cache folder"));
+
+            if (filteredLogOutput.Count() == 1 &&
+                filteredLogOutput.FirstOrDefault().Contains("Shutting down tenant \"Default\" because of idle timeout"))
```
wAsnk: This shouldn't be an issue since it's just an info log, it shouldn't cause the UI test to fail. Is this really needed?

test/Lombiq.OSOCE.Tests.UI/Helpers/AssertAppLogsHelpers.cs
```diff
@@ -0,0 +1,63 @@
...
+            // Temporarily filtering out irrelevant cache errors from logs.
+            var filteredLogOutput = messages.Where(message =>
+                !message.Contains("orchard-log-") &&
+                !message.Contains("Microsoft.Hosting.Lifetime|INFO|") &&
+                !string.IsNullOrEmpty(message) &&
+                !message.Contains("System.IO") &&
+                !message.Contains("|WARN|") &&
+                !message.Contains("|Lombiq.TrainingDemo.Services.DemoBackgroundTask|ERROR|Expected non-error") &&
+                !message.Contains("Azure Media Storage is enabled but not active because the 'ContainerName' is missing or empty") &&
+                !message.Contains("Azure Media Storage is enabled but not active because the 'ConnectionString' is missing") &&
```
wAsnk: Since you have these set in the appsettings.development.json, do you still need it?

test/Lombiq.OSOCE.Tests.UI/Helpers/AssertAppLogsHelpers.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            var messages = logOutput.SplitByNewLines().ToList();
+
+            // Temporarily filtering out irrelevant cache errors from logs.
+            var filteredLogOutput = messages.Where(message =>
+                !message.Contains("orchard-log-") &&
+                !message.Contains("Microsoft.Hosting.Lifetime|INFO|") &&
+                !string.IsNullOrEmpty(message) &&
+                !message.Contains("System.IO") &&
+                !message.Contains("|WARN|") &&
```
wAsnk: Where did these come from?

test/Lombiq.OSOCE.Tests.UI/UITestBase.cs
```diff
@@ -43,7 +44,7 @@ protected UITestBase(ITestOutputHelper testOutputHelper)
                 configuration.BrowserConfiguration.Headless =
                     TestConfigurationManager.GetBoolConfiguration("BrowserConfiguration:Headless", defaultValue: false);
 
-                configuration.AssertAppLogsAsync = AssertAppLogsDefaultOSOCEAsync;
+                configuration.AssertAppLogsAsync = AssertAppLogsHelpers.AssertOsoceAppLogsAreEmptyAsync;
```
wAsnk, line 47: So you are not using the AssertAppLogsDefaultOSOCEAsync anymore, if so, you can delete that.

wAsnk (changes requested):

src/Lombiq.OSOCE.Web/appsettings.json
porgabi: Without this addition, the UI testing toolbox project's own UI tests start failing (logs filled with errors caused by lack of these strings), presumably because those are run in production mode or something. After adding these strings to the non-development appsettings too, the tests started passing.

Does it cause issues to have these here as well?

test/Lombiq.OSOCE.Tests.UI/UITestBase.cs
porgabi: It's still used in some tests. I think VS does not show the references, but you can find them by searching for `AssertAppLogsDefaultOSOCEAsync`.

test/Lombiq.OSOCE.Tests.UI/Helpers/AssertAppLogsHelpers.cs
porgabi: So these were always in the logs, but logs assertion in `AssertAppLogsDefaultOSOCEAsync` only made sure there are no `|ERROR|` strings in the logs. As I replaced that method with this one (temporarily, mind, as the connection string errors should be gone after Azure Media Storage is not enabled in local development), this is how I filtered them instead.

src/Lombiq.OSOCE.Web/appsettings.json
wAsnk: Oh right, this is OSOCE... You don't really need a development appsettings. You can have everything in the appsettings.json. Please delete the appsettings.development.json and leave the settings in the appsettings.json.

test/Lombiq.OSOCE.Tests.UI/Helpers/AssertAppLogsHelpers.cs
porgabi: It causes the test to fail as in line 48 `filteredLogOutPut` should be empty. This only concerns a singular test (`ShuttingDownIdleTenantsShouldWork`), which has its own app logs assertion method, but it's overridden by `AssertOsoceAppLogsAreEmptyAsync`. I found it simpler to temporarily add this here while the new method is necessary for other tests.

test/Lombiq.OSOCE.Tests.UI/Helpers/AssertAppLogsHelpers.cs
porgabi: I think some tests still had the connection string related errors in the logs even though I added them to the development appsettings. An alternative is to add the connection strings to the non-development appsettings in this solution too, I suppose.

EDIT -- Actually, this solution already has the production appsettings additions. So I'm not certain what caused those to still appear in the logs.

test/Lombiq.OSOCE.Tests.UI/Helpers/AssertAppLogsHelpers.cs
wAsnk: Hmm that is also a solution, but the problem is that it will fail on any new warning or info log so you have to add everything here later on. I guess no one really wants to do that. It's better to use the general approach where you filter those errors out that are not actually errors and then check:
```
logOutput.ShouldNotContain("|ERROR|");
logOutput.ShouldNotContain("|FATAL|");
```

Also because of the check on line 25 you actually will never have a case where line 53 will be true and it would fail on 55 and 56. It can't happen because the log will be empty when you get there.

porgabi: You're right, I'll change it so it checks for "error" and "fatal" instead.

I'm counting on this entire method being unnecessary once it's clarified how and when Azure Media Storage should be enabled in development.

test/Lombiq.OSOCE.Tests.UI/Helpers/AssertAppLogsHelpers.cs
porgabi: Removed it as it's now checking for "error" and "fatal" instead.

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Helpers/AssertAppLogsHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        var logOutput = await webApplicationInstance.GetLogOutputAsync(cancellationToken);
+
+        if (!string.IsNullOrEmpty(logOutput))
+        {
+            var messages = logOutput.SplitByNewLines().ToList();
+
+            // Temporarily filtering out irrelevant cache errors from logs.
+            var filteredLogOutput = messages.Where(message =>
+                !message.Contains("orchard-log-") &&
```
wAsnk: What is the error that requires this line?

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Helpers/AssertAppLogsHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+                !string.IsNullOrEmpty(message) &&
+                !message.Contains("|Lombiq.TrainingDemo.Services.DemoBackgroundTask|ERROR|Expected non-error") &&
+                !message.Contains("Azure Media Storage is enabled but not active because the 'ContainerName' is missing or empty") &&
+                !message.Contains("Azure Media Storage is enabled but not active because the 'ConnectionString' is missing") &&
+                !message.Contains("OrchardCore.Media.Core.DefaultMediaFileStoreCacheFileProvider.TryDeleteDirectoryAsync") &&
```
wAsnk: I guess this is info only as well

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Helpers/AssertAppLogsHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+            filteredLogOutput.ShouldNotContain("|FATAL|");
+
+            return;
+        }
+
+        if (canContainWarnings)
+        {
+            logOutput.ShouldNotContain("|ERROR|");
+            logOutput.ShouldNotContain("|FATAL|");
+        }
```
wAsnk: You still return on line 42 if the log had anything in it. So you'll get here if it is empty, thus this really doesn't do anything.

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Helpers/AssertAppLogsHelpers.cs
porgabi: The log output string always begins with a line that contains this text.

wAsnk: Okay but you don't really need to filter that out if you are checking for errors and fatals.

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Helpers/AssertAppLogsHelpers.cs
porgabi: Number 1 and number 4 are `ERROR`, number 2 and number 3 are `FATAL`.

image

image

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Helpers/AssertAppLogsHelpers.cs
porgabi: True!

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Helpers/AssertAppLogsHelpers.cs
wAsnk: The error is on "OrchardCore.Media.Core.DefaultMediaFileStoreCacheFileProvider|ERROR|Error deleting cache folder"
but not on "OrchardCore.Media.Core.DefaultMediaFileStoreCacheFileProvider.TryDeleteDirectoryAsync", right?

But why azure data protection errors are still there? :/

porgabi: > The error is on "OrchardCore.Media.Core.DefaultMediaFileStoreCacheFileProvider|ERROR|Error deleting cache folder"
but not on "OrchardCore.Media.Core.DefaultMediaFileStoreCacheFileProvider.TryDeleteDirectoryAsync", right?

Yeah, they are the same error but on different lines, so I'll remove the filtering for `TryDeleteDirectoryAsync`.

> But why azure data protection errors are still there? :/

I'll try a test run without these to see if they are still there (or if I just saw it wrong earlier).

porgabi: > I'll try a test run without these to see if they are still there (or if I just saw it wrong earlier).

Seems it was just a case of me being confused by the errors.

wAsnk (approved):

PR #224, opened 2022-10-10
OSOE-84: Fixing dotnet restore

0liver:
OSOE-84

Trying to get to the bottom of the broken `dotnet restore` call.


0liver:
This has become a dead end. Please go to #225.

PR #229, opened 2022-10-13
SNOW-147: Fix SafeJsonAsync and add tests

sarahelsaig:
SNOW-147


test/Lombiq.OSOCE.Tests.UI/Tests/BehaviorHelpfulLibrariesTests.cs
```diff
@@ -0,0 +1,78 @@
...
+
+public class BehaviorHelpfulLibrariesTests : UITestBase
+{
+    public BehaviorHelpfulLibrariesTests(ITestOutputHelper testOutputHelper)
+        : base(testOutputHelper)
+    {
+    }
+
+    [Theory, Chrome]
+    public Task SafeJsonShouldWork(Browser browser) =>
```
Piedone: Shouldn't this rather be a unit test?

sarahelsaig: The actually important is part [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/blob/81e2073ef1175e390f814aa5f296116954a862bc/test/Lombiq.OSOCE.Tests.UI/Tests/BehaviorHelpfulLibrariesTests.cs#L73-L75) and I don't think I could do that satisfactorily in unit test. My main goal was to ensure that the errors get logged because they weren't previously unless fatal. The rest is just there for completion's sake.

Piedone: You can test that in a quite straightforward way with a mocked `ILogger`.

sarahelsaig: At that point I'd have to create a controller with mocked `HttpContext` and a `IServiceProvider` in it. Also set up dependency injection for `ILoggerFactory` and `IStringLocalizerFactory` so almost everything would be synthetic and quite far away from the real method. Is this good? Seems pretty unappealing to me.

Piedone: Can`t you just test `SafeJsonAsync`?

sarahelsaig: That's what I'm talking about. What I have listed are all _inside_ `SafeJsonAsync`. It's an extension method for `Controller` and it uses `Controller.HttpContext.RequestServices` to request the mentioned factory services to create `ILogger<T>` and `IStringLocalizer<T>` where `T` is the type of the specific caller so the log is still associated with the caller as normally expected.

Piedone: Hmm, I see. Mocking everything necessary there indeed looks like a pain, but still, it seems to me that a more lightweight unit test is more appropriate (after all, we're only testing code-level behavior here, and for a single method). `Lombiq.Tests` might have some suitable helpers or if you develop those, please add them there.

BTW `SafeJsonShouldWork` also should rather be implemented by using a UI test extension method from HL like the other such UI tests.

sarahelsaig: Your comment is confusing. To put it simply, are you telling me "yes it's a pain but do it anyway"?

Piedone: It's a pain to create the test, but after that, we get rid of the pain of having to run and maintain a UI test that just tests a single method in the end.

sarahelsaig: I did it and it was even more painful than expected.

Piedone: Think about the friends you made along the way.

sarahelsaig: Sure, I will write about it in my special diary.

image

sarahelsaig:
"Verify that .NET packages are consolidated" keeps failing since I've merged from `dev`. Anyone knows what's going on?

test/Lombiq.OSOCE.Tests.UI/Tests/BehaviorHelpfulLibrariesTests.cs
Piedone: :D

Piedone:
The build output shows where the issue is. You updated packages in HL that other projects use too, so NuGet packages should now be consolidated.

sarahelsaig:
I did what? I didn't update packages, just merged from dev.

sarahelsaig:
I see the problem: I've merged from dev after ea3f555786678bfb2dba547cf3c2c4bea45d07b0 but before f346024a7fbd4e9a7c8edf6a67d049da05c20f18. What bad luck!

Psichorex (approved):

PR #243, opened 2022-10-21
OSOE-409: Speed up Tests execution step ⭐

0liver:
OSOE-409


0liver:
Thanks for the submodule update 😏

PR #222, opened 2022-10-04
OSOE-156: Code styling

Piedone:
OSOE-156

PR #266, opened 2022-11-07
OSOE-423: Offer Media management testing as part of TestBasicOrchardFeaturesAsync() too in Lombiq.UITestingToolbox

Luko6:
[OSOE-423](https://lombiq.atlassian.net/browse/OSOE-423)

Fixes #222


Piedone (changes requested):
Also update the NuGetTest solution/recipe, since there's also a `BasicOrchardFeaturesShouldWork` there.

Piedone (approved):

Piedone (approved):

PR #344, opened 2022-12-28
OSOE-518: Fix formatting of action/workflow files in Lombiq.GitHub.Actions

BenedekFarkas:
[OSOE-518](https://lombiq.atlassian.net/browse/OSOE-518)


Piedone (changes requested):
Branch references are missing in the workflows here

BenedekFarkas:
> Branch references are missing in the workflows here

Added more!

Piedone (approved):

PR #384, opened 2023-01-30
OSOE-505: Changing the condition against the pull_request event

dorraj07:


wAsnk:
The "Add Jira Issue Code to Pull Request" did not run at all, but should've.

PR #403, opened 2023-02-13
OSOE-536: In NuGet publish workflow, restrict valid versions to vM.N.O or v.M.N.O-alpha.X.[issue-code]

adelgeiszt:
[OSOE-536](https://lombiq.atlassian.net/browse/OSOE-536)

[OSOE-536]: https://lombiq.atlassian.net/browse/OSOE-536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -39,7 +39,6 @@ jobs:
       machine-types: "['ubuntu-22.04', 'windows-2022']"
       build-directory: NuGetTest
       timeout-minutes: 20
-      build-create-binary-log: "true"
 
   spelling:
     name: Spelling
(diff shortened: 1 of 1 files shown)


adelgeiszt:
@0liver The task is done, however, I'm getting these Build and Test - NuGetTest solution workflow fails, caused by `ERROR: TypeError: Cannot read properties of undefined` and after looking into it, it seems to me that it's unrelated to my solution.

0liver:
The first thing I would try is to re-run the failing workflow:

image

If the error persists, look into it a bit deeper.

0liver:
Wow the build passed now!

While I'm happy for you, it's also not nice to have it fail randomly. But that's not your problem 😁

0liver (approved):

PR #405, opened 2023-02-16
OSOE-476: Extract inline javascript block into file, and register for usage in Lombiq.HelpfulExtensions

TheHydes:
[OSOE-476](https://lombiq.atlassian.net/browse/OSOE-476)

[OSOE-476]: https://lombiq.atlassian.net/browse/OSOE-476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -39,7 +39,6 @@ jobs:
       machine-types: "['ubuntu-22.04', 'windows-2022']"
       build-directory: NuGetTest
       timeout-minutes: 20
-      build-create-binary-log: "true"
 
   spelling:
     name: Spelling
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -39,6 +39,7 @@ jobs:
       machine-types: "['ubuntu-22.04', 'windows-2022']"
       build-directory: NuGetTest
       timeout-minutes: 20
+      build-create-binary-log: "true"
```
DemeSzabolcs: This was just for debugging if I know it right. So you shouldn't need this change anymore.

DemeSzabolcs (changes requested):

DemeSzabolcs (approved):

DemeSzabolcs (approved):

PR #480, opened 2023-05-25
NEST-391: Maintenance to make a role site owner

barthamark:
[NEST-391](https://lombiq.atlassian.net/browse/NEST-391)

[NEST-391]: https://lombiq.atlassian.net/browse/NEST-391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


NuGetTest/src/Lombiq.OSOCE.NuGet.Web/Lombiq.OSOCE.NuGet.Web.csproj
```diff
@@ -36,7 +36,7 @@
     <PackageReference Include="Lombiq.Hosting.Tenants.EnvironmentRobots" Version="4.0.3-alpha.19.osoe-621" />
     <PackageReference Include="Lombiq.Hosting.Tenants.FeaturesGuard" Version="4.0.2-alpha.1.nest-422" />
     <PackageReference Include="Lombiq.Hosting.Tenants.IdleTenantManagement" Version="4.0.2-alpha.1.nest-422" />
-    <PackageReference Include="Lombiq.Hosting.Tenants.Maintenance" Version="4.0.3-alpha.16.nest-326" />
+    <PackageReference Include="Lombiq.Hosting.Tenants.Maintenance" Version="4.0.3-alpha.1.nest-391" />
```
wAsnk: Please also update the other Lombiq.Hosting.Tenants NuGet versions. Thank you. 🍓

wAsnk (changes requested):

NuGetTest/src/Lombiq.OSOCE.NuGet.Web/Lombiq.OSOCE.NuGet.Web.csproj
0liver: You should keep increasing the number after the `4.0.3-alpha` part to simplify updating to a newer version.

barthamark: It'll become unmanageable with parallel development so I wouldn't increase the number after the `alpha`. The issue number differentiates it already.

barthamark: Apparently, we are on 4.1.0 in dev already, so it'll change anyway.

0liver: > It'll become unmanageable with parallel development so I wouldn't increase the number after the `alpha`. The issue number differentiates it already.

The idea behind this versioning scheme is to be able to simply increment the number behind the `alpha`, no matter on which issue you are currently working. This allows for a very pleasing experience with the NuGet package manager from VS where you can simply update to all latest versions, including pre-release ones.

There is the edge-case, you are correct, that two devs work on the same module in parallel and update the version at the same time or close to each other, but that happens extremely rarely and doesn't add any further harm, so I think we should be able to accept it.

PR #486, opened 2023-06-03
OSOE-464: Executing dotnet test with process timeout

dministro:
OSOE-464


sarahelsaig (approved):

sarahelsaig:
@dministro Can you confirm if the other examples of `blame-hang-timeout` should or shouldn't be updated, as mentioned in the review comments above?

dministro:
> @dministro Can you confirm if the other examples of `blame-hang-timeout` should or shouldn't be updated, as mentioned in the review comments above?

Good question, but let's use the same config everywhere.

PR #517, opened 2023-07-02
OSOE-517: Automate managing issue branch references in workflow files

davidpuplava:
[OSOE-517](https://lombiq.atlassian.net/browse/OSOE-517)

[OSOE-517]: https://lombiq.atlassian.net/browse/OSOE-517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DemeSzabolcs:
Hello David, I will review your issue from Lombiq, sorry for the delay!

I think Lombiq.GitHub.Actions is not on the latest commit of the OSOE-517 branch in the submodule, since [validate-resusable-action-workflow-refs.yml](https://github.com/Lombiq/GitHub-Actions/compare/bf8e26ed06752c73e862f573ad22d5d15b82c9a5...16680885dd0bf3ea5a9e21d257efa9c5f44f4a18#diff-9cdde25a29f521a006e61d65c4e82dd50b8c88be91379c86b9b1f83068da98c7)
is not present there: https://github.com/Lombiq/GitHub-Actions/pull/243/files

davidpuplava:
Hi Szabolcs - no problem, thank you for reviewing! You are correct, I forgot to update the submodule reference to the latest commit. I see your feedback comments in the submodule repo, so I'll make the changes there and then commit the latest reference when complete. Thanks again!

davidpuplava:
Hi @DemeSzabolcs - I pushed the latest code changes for this. The new changes include a new workflow that triggers based on a pull request. I've added two new actions. One to get a list of all file changes from the PR (a git diff between PR branch and PR target branch). Another to determine the list of Workflows and Actions that have been modified (from the list of git diff files). 

The list of modified GHA items is then passed into the original `verify-gha-refs` action as the list of "called base repos" which is how we determine if the `@ref` should be checked. If the called workflow/action hasn't changed it doesn't get flagged. If it has changed, it will get flagged.

Here are the following use cases I think this will cover:
- on pull request
  - edit the README.md file, workflow passes because no GHA actions/workflows have changed
  - edit any workflow/action .yaml (or .yml) file, see that any places that calls the workflow (i.e. `uses: Lombiq/...` will get flagged
  - edit any GHA action directory file such as a `*.ps1` or `*.txt` file, notice that the GHA action is flagged if it is used anywhere
- on push to `dev` branch
  - will check all references and flag any that do not match

Note that the `verify-this-gha-refs` workflow includes the `issue/OSOE-517` on push which is only temporary for the purposes of reviewing this PR.

I did not create scripts to automatically update the references. I found it was easier to just find/replace using VSCode or Visual Studio. These scripts would be more useful for a second iteration of this feature to automatically update the references as part of a workflow.

DemeSzabolcs:
Hello @davidpuplava, thanks for the in-detail description!

Currently there are no file changes in this PR 
image

I guess you switched branches.

I can see an example for this
> on push to dev branch
will check all references and flag any that do not match

as you mentioned here

> Note that the verify-this-gha-refs workflow includes the issue/OSOE-517 on push which is only temporary for the purposes of reviewing this PR.

Could you please also show an example for this, in your PR? (so for example making it to fail)
> on pull request
edit the README.md file, workflow passes because no GHA actions/workflows have changed
edit any workflow/action .yaml (or .yml) file, see that any places that calls the workflow (i.e. uses: Lombiq/... will get flagged
edit any GHA action directory file such as a *.ps1 or *.txt file, notice that the GHA action is flagged if it is used anywhere

davidpuplava:
Hi @DemeSzabolcs

Regarding
> Could you please also show an example for this, in your PR? (so for example making it to fail)

Yes, please take a look at [this action run](https://github.com/Lombiq/GitHub-Actions/actions/runs/5734506893). It shows three places where I should update the `@dev` ref to `@issue/OSOE-517`. 

- The first two are calls to the the spelling action because I modified the `excludes.txt` file in that action directory. 
- The last one is because I modified the `validate-submodule-pull-request` workflow, so the `validate-this-pull-request` should change because it calls the workflow that I changed.

DemeSzabolcs:
Hello @davidpuplava!

Thanks, it looks good in general. Just two things I noticed:

1. PowerShell Static Code Analysis is failing (in this PR), so you will need to correct the problematic files: 
image

2. There is a situation where the `Verify Changed Workflow Files Match` step will fail, but it shouldn't. Let me explain:

Somebody changes a file called `test.yml`, they also change the references to the issue branch `issue/xxx-123`. Let's say it's done and the reviewer comes. The reviewer accepts the PR and will change the references back to `dev` since they want to merge it into `dev`.
However at this point `Verify Changed Workflow Files Match` will fail, because `test.yml` is changed, but the references are no longer using the issue branch, which is logical at this point, since the reviewer is ready for the merge.

To fix this my first thought (though it might be not the best solution), is to use some kind of a label that would be added initially, and removed when the reviewer accepts the PR. Then this label could be detected by that action and the action could be skipped if the label is no longer present. Or maybe we can even detect the acceptance of a reviewer without labels.

What do you think? Is the situation that I described correct?

davidpuplava:
Thanks @DemeSzabolcs 
>PowerShell Static Code Analysis is failing (in this PR), so you will need to correct the problematic files

Yes, I will get these errors fixed and push a new change for the PR.

> There is a situation where the Verify Changed Workflow Files Match step will fail, but it shouldn't. Let me explain

Yes, I think to narrow the conditions for when the `Verify Changed Workflow Files Match` should run makes sense. I'm not very knowledgeable on how the labels work in GitHub but I think I can see how other LGHA workflows and actions use them. Alternatively, I wonder if we can change the job to skip when the PR is approved. Or better yet, run the job in a way that ensures all the refs point back to the default branch when the PR is in the approved state. I found some details [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-when-a-pull-request-is-approved).

Let me do a little thinking on it today, and I'll respond with what I find.

--EDIT--
@DemeSzabolcs Are reviews required for PRs into the `dev` branch? If so, do you have some kind of 'approve' button? If so, we can use the `pull_request_review` event trigger and check to see if the PR was approved to skip the check. If not, then the label solution you suggested may be the best option.

DemeSzabolcs:
@davidpuplava Sorry, I didn't see the comment edit at first, there was no notification about that.

We don't have a separate "approve" button, we only use the GH one: 
image

And it's displayed like this: 
image

I'm not sure, but to me it would be logical if we could get this data (if the PR is approved or not) from GH (but there might be no way of doing that).

davidpuplava:
Hi @DemeSzabolcs - Sorry about the comment EDIT, I'll add a new reply next time.
> We don't have a separate "approve" button, we only use the GH one:

Great! I think this is exactly what we need to get the `pull_request_review` event trigger to work with something like `if: github.event.review.state == 'approved'`. I will try to push something new today or tomorrow.

>I'm not sure, but to me it would be logical if we could get this data (if the PR is approved or not) from GH (but there might be no way of doing that).

Agreed that using this data is logical. My only concern is what the developer/reviewer experience is like. My initial thought is that any pushes after approval will check the `@ref` is for the target branch rather than the PR branch. I'll see what I can come up with.

davidpuplava:
@DemeSzabolcs almost ready. I've ended up with a new GitHub action that will check a pull requests reviews and ensure they're all in the state APPROVED, which I can then use to inform the validate-pull-request job. I just need to test in my private repo but need to figure out the GH_TOKEN to use the `gh api` command to grab the PR reviews. I'll likely have tomorrow.

davidpuplava:
Hi @DemeSzabolcs - I believe this now addresses the issue you described
> There is a situation where the Verify Changed Workflow Files Match step will fail, but it shouldn't. Let me explain

I created a new [check-pull-request-reviews](https://github.com/Lombiq/GitHub-Actions/tree/issue/OSOE-517/.github/actions/check-pull-request-reviews) action that will evaluate the latest PR review and determine if it is an approved review (using the GitHub CLI).

I modified the [validate-this-gha-refs-pr](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-this-gha-refs-pr.yml) workflow to use this new `check-pull-request-review` action. I added and new [Determine Expected Ref for GHA Files](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-this-gha-refs-pr.yml#L57) step which will decide if the issue branch (e.g. `@issue/OSOE-517`) or the target merge branch (e.g. `@dev`) name should be checked. If the PR is not yet approved, it will check the issue branch. If the PR is approved, it will check for the target base branch.

I also modified the [validate-this-gha-refs](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-this-gha-refs.yml#L7) to trigger when a PR review is submitted (such as when a PR is approved). Notice that there is a new conditional step [here](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-this-gha-refs.yml#L21) to run with the base branch on approval. This will make sure that the PR has all the references changed back to dev when a PR is approved.

Please let me know if you have any questions.

DemeSzabolcs:
Hello @davidpuplava! Thank you, I will do a review in the upcoming days.

DemeSzabolcs:
@davidpuplava I added some comments in the submodule. The OSOCE PR itself (this) is okay. It's failing for an unrelated reason, it's the same on [dev](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/5801147267/job/15802216327), but it's unrelated for this issue.

DemeSzabolcs (approved):

Piedone:
Please merge from `dev` to fix the build error, see: https://github.com/Lombiq/Orchard-Base-Theme/issues/84

DemeSzabolcs (approved):

DemeSzabolcs (approved):

DemeSzabolcs (approved):

DemeSzabolcs (approved):

PR #526, opened 2023-07-17
OSOE-603: ⭐ Upgrade to pnpm@8

sarahelsaig:
[OSOE-603](https://lombiq.atlassian.net/browse/OSOE-603)

[OSOE-603]: https://lombiq.atlassian.net/browse/OSOE-603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/build-and-test.yml
```diff
@@ -20,6 +20,7 @@ jobs:
       ui-test-parallelism: 0
       build-create-binary-log: "true"
       blame-hang-timeout: "5m"
+      build-verbosity: normal
```
Piedone: Isn't the default suitable?

package.json
```diff
@@ -1,11 +1,11 @@
 {
   "private": true,
   "devDependencies": {
-    "eslint": "^8.25.0",
+    "eslint": "^8.45.0",
```
Piedone: And update would be great, but all the other package.jsons are 8.25.0.

Piedone (changes requested):

.github/workflows/build-and-test.yml
sarahelsaig: It was just diagnostics, I removed most but I guess forgot this one.

package.json
sarahelsaig: It wasn't intentional, not sure why but sometimes it tries to update it despite the lock. Reverted.

Piedone (approved):

pnpm-lock.yaml
Piedone, line 1: A rebuild causes this for me:

image

So something apparently needs to be done with the package.json file.

Piedone (changes requested):

pnpm-lock.yaml
sarahelsaig: That often happens for me, with this or that NE-using project. I usually just revert it.

Piedone: We shouldn't have to revert it. This, and only this file, happens on every rebuild for me.

sarahelsaig: Yeah, it do be like that. it's not part of this issue, but I will investigate soon.

Piedone:
Thank you!

--
Zoltán Lehóczky
Co-founder and managing director
Lombiq Technologies Ltd.
https://lombiq.com
[image]<https://lombiq.com/>
Please note that we store your personal data in accordance with our privacy policy<https://lombiq.com/privacy-policy>, but you can anytime choose to withdraw your previous consent by writing to ***@***.***
________________________________

Piedone (approved):

Piedone (approved):

PR #602, opened 2023-10-31
OSOE-430: Fix PNPM link error and reapply #578

sarahelsaig:
[OSOE-430](https://lombiq.atlassian.net/browse/OSOE-430)

re #578

[OSOE-430]: https://lombiq.atlassian.net/browse/OSOE-430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sarahelsaig:
There is nothing new in this branch or Lombiq.VueJs compared to the last PR. Lombiq.NodeJs.Extensions has relevant changes here:
https://github.com/Lombiq/NodeJs-Extensions/pull/78/files/43ec6acf4cc6b61c21bb37421852fa97be9a956f..8ebcad27839d93abf6a35d35cf802adfeb5bcb6f

sarahelsaig:
So this was about the `pnpm link` error that appeared on Windows after the original PR was merged. In the end I got around it by only applying the modified linker command to non-Windows. I retried this a bunch of times today, and the problem does not arise any more.

src/Libraries/Lombiq.HelpfulLibraries
dministro, line 1: I think this coming from the revert commit. Please update submodules to the latest. Here and where needed.

dministro (changes requested):

src/Libraries/Lombiq.HelpfulLibraries
sarahelsaig: It's not related to the revert. It's coming from updating all submodules using from CLI like this:
```bash
git submodule foreach 'git checkout "issue/OSOE-430-newer" || git checkout dev && git pull'
git submodule foreach 'git merge "origin/dev"'
```
Apparently HL wasn't at the latest tip in OSOCE's dev branch the moment I merged from dev the last time.

sarahelsaig:
I’ve fixed the error handling that caused this exception: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/7603089215/job/20704395923?pr=602#step:6:144

But there is an underlying problem:

Untitled
> TypeError: Cannot use ‘in’ operator to search for ‘line’ in `PNPM is not installed. Please check the prerequisites for Lombiq Node.js Extensions at https://github.com/Lombiq/NodeJs-Extensions#prerequisites` .

This is all related to [this code](https://github.com/Lombiq/NodeJs-Extensions/blob/dev/Lombiq.NodeJs.Extensions/scripts/run-pnpm-script.js#L12-L17):

```js
// Check if pnpm is installed.
try { execSync('pnpm -v'); }
catch (_) {
    panic('PNPM is not installed. Please check the prerequisites for Lombiq Node.js Extensions at ' +
        'https://github.com/Lombiq/NodeJs-Extensions#prerequisites');
}
```

It’s the very first thing run-pnpm-script.js does, so we can be certain the “Access is denied.” is the output of the `pnpm -v`. Just a few steps prior in the same build you can see that PNPM was successfully activated.
Untitled

The target that installs PNPM has [this comment](https://github.com/Lombiq/NodeJs-Extensions/blob/dev/Lombiq.NodeJs.Extensions/build/NodeJsExtensionsNpmPackage.targets#L6) which reinforces my first guess that this is a parallelization problem. but shouldn't the mutex prevent that? @0liver do you have any input perhaps?

0liver:
I've looked at the info you've posted, and I don't have any good idea why the `pnpm -v` command would fail _after_ the successful PNPM activation, especially given that the command had to wait for at least 2.7 seconds before starting:

image

I'll try to take another round of thinking this through, but for that's all I can say.

dministro (approved):

PR #603, opened 2023-10-31
NEST-490: Double-logging in Application Insights for the same idle shutdown event

wAsnk:
[NEST-490](https://lombiq.atlassian.net/browse/NEST-490)

[NEST-490]: https://lombiq.atlassian.net/browse/NEST-490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Lombiq.OSOCE.Web/Program.cs
```diff
@@ -16,6 +16,7 @@
 // create UI tests check out the project.
 builder.Services
     .AddSingleton(configuration)
+    .AddOrchardCoreApplicationInsightsTelemetry(configuration)
```
Piedone: This causes the following empty telemetry to be attempted to be sent:

image

This should rather be commented out by default with a comment, but enabled during UI testing but only for `ApplicationInsightsTrackingInOfflineOperationShouldWorkAsync()` (you can e.g. detect the `EnableOfflineOperation` config set there).

src/Lombiq.OSOCE.Web/Program.cs
Piedone, line 1: In the appsettings file the `ApplicationInsights` section should rather be in the root, right?

Piedone (changes requested):

src/Lombiq.OSOCE.Web/Program.cs
wAsnk: Yeah it should be so it is unified. But the config would be picked up from here as well.

src/Lombiq.OSOCE.Web/Program.cs
wAsnk: Hmm you are right it should be restricted somehow, but actually it would be better to add a check within the module, so it is only initiated when it is needed. Then the empty requests won't be sent.

src/Lombiq.OSOCE.Web/Program.cs
Piedone: Please move it to the root, then.

src/Lombiq.OSOCE.Web/Program.cs
Piedone: Even better, yes.

src/Lombiq.OSOCE.Web/Program.cs
wAsnk: Yup, forgot to commit it.

Piedone (approved):

Piedone (approved):

Piedone (approved):

PR #628, opened 2023-11-24
OSOE-729: Prevent Azure-related workflows running at the same time when interacting with the same environment(s)

BenedekFarkas:


Psichorex (approved):

Psichorex (approved):

Psichorex:
Seems like the same bug is happening here as described in https://github.com/Lombiq/UI-Testing-Toolbox/issues/321.
Shall we neglect this here? @BenedekFarkas
Edit: Added a re-run because it should pass after it.

BenedekFarkas:
Yep, we're good, that's a flaky test (and unrelated to these changes anyway). I've added one more tiny update, it will be merged automatically if the checks are all green.

PR #637, opened 2023-12-05
OSOE-481: Action pulls latest SQL Server (=2022-latest) but 2019-latest is used in Lombiq.GitHub.Actions

Psichorex:
OSOE-481


Psichorex:
@Piedone A UI test is failing most probably because the docker now pulls and uses sql-server-2022 but 
https://github.com/Lombiq/UI-Testing-Toolbox/blob/88ec8caf6940e7f1445fe4993701fc0ce5ce5e48/Lombiq.Tests.UI/Services/SqlServerManager.cs#L173
This line receives sql2019 as a container name
This is coming from the error message of the failed docker command from Cli.Wrap()
`Command: docker cp sql2019:/data/Snapshots/Database.bak /home/runner/actions-runner/_work/Open-Source-Orchard-Core-Extensions/Open-Source-Orchard-Core-Extensions/test/Lombiq.UITestingToolbox/Lombiq.Tests.UI.Samples/bin/Debug/net6.0/Temp/SetupSnapshot/SqlServer-57163436/Database.bak`
Do you know maybe where it is exactly set what is the docker image name that the linux env will use? Because I wasn't able to find it yet.

Piedone:
I don't know from the top of my head, I'd need to run it with the debugger and figure out from the call stack too.

dministro (changes requested):
Please merge from dev

dministro (approved):

dministro (changes requested):
Next time if you change something Windows related please adjust `build-and-test-windows` workflow accordingly.

The Windows workflow is failing. See: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/7160130952/job/19494163261#step:8:95

Psichorex:
> Next time if you change something Windows related please adjust `build-and-test-windows` workflow accordingly.
> 
> The Windows workflow is failing. See: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/7160130952/job/19494163261#step:8:95

@dministro As far as I see it's not failing because of I didn't change `build-and-test-windows` because that only has references for other workflows and if I had to change the references there `verify-gha-refs` should have caught that.

I changed the refs though and it is still failing. 
image
It's the `Wait-SqlServer` that is failing after according to logs sqlserver2022 being successfully installed.

Psichorex:
Yes so I switched back to `sql-server-express` here:
image
And the CI is not failing anymore. Any ideas what could be causing this?

dministro:
> Yes so I switched back to `sql-server-express` here: image And the CI is not failing anymore. Any ideas what could be causing this?

Did you tried to install and run the SqlServer locally using choco?

Psichorex:
> > Yes so I switched back to `sql-server-express` here: image And the CI is not failing anymore. Any ideas what could be causing this?
> 
> Did you tried to install and run the SqlServer locally using choco?

I tried to install with both commands but `sql-server-2022` fail for some reason. That my be a local issue but `sql-server-express` worked for the first time. So at this point again I don't know what exactly causes this to fail because the logs show that sql-server-2022 command did install the instance.

Psichorex:
@Piedone I can't really determine why the `sql-server-2022` command is failing and why `sql-server-express` is working without anything else being changed. We could revert to `sql-server-express` and maybe extract this into another issue to investigate whether the other command is failing. I didn't find anything in the documentation that would indicate a dependency or anything.

Piedone:
So do you mean that we'd have SQL Server Express 2022 (it's important that it'll also be v2022) installed instead of SQL Server Developer 2022? I'm OK that, but Jocó?

Psichorex:
> So do you mean that we'd have SQL Server Express 2022 (it's important that it'll also be v2022) installed instead of SQL Server Developer 2022? I'm OK that, but Jocó?

This command `choco install sql-server-express` is dedicated for SQL SERVER EXPRESS 2022 according to the docs. So it is surely a 2022 version.

Piedone:
That's because that's the latest version currently. Also use `--version` like they show for 2019 [here](https://community.chocolatey.org/packages/sql-server-express/2019.20200409).

dministro:
> I can't really determine why the `sql-server-2022` command is failing and why `sql-server-express` is working without anything else being changed. 

@Psichorex Did you tried to use `(local)` or `.` as server name without `\SQLEXPRESS` in connection parameters/strings? IfI remember well the `-S` command line parameter default value is `.` for `sqlcmd` so you can simple remove it from the parameter list.
The `SQLServer Developer Edition` creates an instance with a name something like `MSSQLSERVER` which is the default so you don't have to specify the instance name in the server name. This means the instance with name `SQLEXPRESS` does not exists which can be the root cause of the issue.

> We could revert to `sql-server-express` and maybe extract this into another issue to investigate whether the other command is failing. I didn't find anything in the documentation that would indicate a dependency or anything.

If you already tried to adjust connection parameters in `Wait-SqlServer.ps1` and `Invoke-SolutionTests.ps1` like above and doesn't work, then please create an issue for it and describe what is the problem here exactly what you have tried, etc.

If you have not tried please give it a try.

Psichorex:
So far I tried these:
`$connectionStringStem = 'Server=.\MSSQLSERVER;Database=LombiqUITestingToolbox_{{id}};Integrated Security=True'`
`$connectionStringStem = 'Server=.\(local);Database=LombiqUITestingToolbox_{{id}};Integrated Security=True'`
And both failed the same way. As for SS2022 Dev as you mentioned the default name should be MSSQLSERVER so if that doesn't I suspect it has to do with something else.

Psichorex:
> > I can't really determine why the `sql-server-2022` command is failing and why `sql-server-express` is working without anything else being changed.
> 
> @Psichorex Did you tried to use `(local)` or `.` as server name without `\SQLEXPRESS` in connection parameters/strings? IfI remember well the `-S` command line parameter default value is `.` for `sqlcmd` so you can simple remove it from the parameter list. The `SQLServer Developer Edition` creates an instance with a name something like `MSSQLSERVER` which is the default so you don't have to specify the instance name in the server name. This means the instance with name `SQLEXPRESS` does not exists which can be the root cause of the issue.
> 
> > We could revert to `sql-server-express` and maybe extract this into another issue to investigate whether the other command is failing. I didn't find anything in the documentation that would indicate a dependency or anything.
> 
> If you already tried to adjust connection parameters in `Wait-SqlServer.ps1` and `Invoke-SolutionTests.ps1` like above and doesn't work, then please create an issue for it and describe what is the problem here exactly what you have tried, etc.
> 
> If you have not tried please give it a try.

I tried like every combination, you can see that in the last commits but non of them seem to be working. So I am reverting for sql-express with version and separate this issue.

dministro:
@Psichorex  Please try `Server=.;` in Windows connection string same like the Linux one. No `Server=.\.;`, `\.` looks weird and can be the root cause of failing tests [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/7261854686/job/19783865013#step:10:2744), please read the log message!

image
image

commit: https://github.com/Lombiq/GitHub-Actions/pull/293/commits/ba19eaa040758e9b8399feaec5983994869c8023

tools/Lombiq.Analyzers.PowerShell
dministro, line 1: Why was this changed?

dministro (changes requested):

tools/Lombiq.Analyzers.PowerShell
Psichorex: image

There is nothing to compare so this is probably a bug. This repo is no branch in my Git Extensions so I didn't touch it.

dministro: @Psichorex Then, please insert the link - where you reported/found this bug - here. Or revert.

Psichorex: @dministro The dev checkout fall behind on my branch most probably due to a pnpm-lock file preventing the updated checkout. I checked out the latest dev in the submodule and the file change disappeared.

dministro (approved):

PR #673, opened 2024-01-18
OSOE-751: Upgrade to Orchard Core 1.8

### LINKED ISSUE #638: Upgrade to Orchard Core 1.8 (OSOE-751)
...once https://github.com/OrchardCMS/OrchardCore/issues/14940 is done.

- https://github.com/Lombiq/Orchard-Core-API-Client/issues/34 is part of this.
- Utilize https://github.com/OrchardCMS/OrchardCore/pull/14759 and https://github.com/OrchardCMS/OrchardCore/pull/14801 where we currently have sync-over-async.
- Perhaps we should do https://github.com/OrchardCMS/OrchardCore/pull/14806 in our forms too?
- Add https://github.com/Lombiq/UI-Testing-Toolbox/issues/324#issuecomment-1827003092 to the release notes of UI testing Toolbox.
- Add a note about the [changed and app-wide setup](https://github.com/Lombiq/Orchard-Azure-Application-Insights#setup-and-basic-configuration) for AI to its release notes.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-751)

Psichorex:
[OSOE-751](https://lombiq.atlassian.net/browse/OSOE-751)
Fixes #638

[OSOE-751]: https://lombiq.atlassian.net/browse/OSOE-751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/Lombiq.OSOCE.Tests.UI/Tests/ModuleTests/BehaviorUIKitShowcaseTests.cs
```diff
@@ -24,7 +19,10 @@ public Task UIKitShowcasePageShouldBeCorrect()
                 async validationResult =>
                 {
                     var errors = (await validationResult.GetErrorsAsync())
-                        .Where(error => !error.ContainsOrdinalIgnoreCase("Prefer to use the native <button> element"));
+                        .Where(error =>
+                        !error.ContainsOrdinalIgnoreCase("Prefer to use the native <button> element") &&
+                        !error.ContainsOrdinalIgnoreCase("<button> must have accessible text") &&
+                        !error.ContainsOrdinalIgnoreCase("Redundant role \"button\" on <button>"));
```
Psichorex, line 32: Reason: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/issues/638#issuecomment-1913566145

test/Lombiq.OSOCE.Tests.UI/Tests/ModuleTests/BehaviorPrivacyTests.cs
```diff
@@ -35,7 +38,14 @@ public async Task ConsentBannerShouldWorkWithRazorAndLiquidBasedThemes()
             context => context.TestConsentBannerWithThemeAsync("TheBlogTheme"));
         // Then should work with Razor-based theme
         await ExecuteTestAfterSetupAsync(
-            context => context.TestConsentBannerWithThemeAsync("TheTheme"));
+            context => context.TestConsentBannerWithThemeAsync("TheTheme"),
+            configuration => configuration.HtmlValidationConfiguration.AssertHtmlValidationResultAsync =
+                async validationResult =>
+                {
+                    var errors = (await validationResult.GetErrorsAsync())
+                        .Where(error => !error.ContainsOrdinalIgnoreCase("Prefer to use the native <button> element"));
```
Psichorex, line 49: Invalid HTML coming from Orchard Core so we have to remove the log.

src/Modules/Lombiq.HelpfulExtensions
Piedone, line 1: This is missing for a lot of projects: https://github.com/OrchardCMS/OrchardCore/pull/14806

test/Lombiq.OSOCE.Tests.UI/Tests/ModuleTests/BehaviorPrivacyTests.cs
Piedone: Please open an issue about this and link it here.

test/Lombiq.OSOCE.Tests.UI/Tests/ModuleTests/BehaviorUIKitShowcaseTests.cs
Piedone: Please open an issue about this and link it here.

test/Lombiq.OSOCE.Tests.UI/Tests/ThemeTests/BlogBaseThemeTests.cs
```diff
@@ -53,6 +54,12 @@ public BlogBehaviorBaseThemeTests(ITestOutputHelper testOutputHelper)
                 var homePageUri = await SetupHelpers.RunBlogSetupAsync(context);
 
                 await context.SignInDirectlyAsync();
+
+                // This is needed to avoid missing jQuery
+                // Can be removed after this issue is resolved and deployed https://github.com/OrchardCMS/OrchardCore/issues/15181
+                await context.GoToFeaturesPageAsync();
+                await context.EnableFeatureDirectlyAsync("Lombiq.HelpfulExtensions.ResourceManagement");
```
Piedone: The decorator doesn't do anything with jQuery anymore.

Piedone (changes requested):

Psichorex:
https://github.com/OrchardCMS/OrchardCore/issues/15222

Piedone:
Always add such issues links for workarounds in the code. And keep to https://lombiq.atlassian.net/wiki/spaces/OCORE/overview#Creating-issues (I added it to our GitHub project now).

src/Modules/Lombiq.HelpfulExtensions
Psichorex: Do we need this on all `<label>` ? Or just in the Base Theme and DataTables where bootstrap is present? Also on admin editors

Piedone: All admin labels and all labels on the frontend where BS is used, but in practice this means all labels.

src/Lombiq.OSOCE.Web/Lombiq.OSOCE.Web.csproj
Piedone, line 1: Upgrade all Microsoft.* packages that are related to a .NET version too. E.g. Microsoft.AspNetCore.Mvc.Testing [has a version for each .NET version](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Testing/), using the same version number as .NET. Such affected packages all do this.

Piedone (changes requested):

src/Lombiq.OSOCE.Web/Lombiq.OSOCE.Web.csproj
Piedone: These remain:

image

Piedone (changes requested):
See https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/673#discussion_r1479051780

src/Lombiq.OSOCE.Web/Lombiq.OSOCE.Web.csproj
Psichorex: I thought only those needed to be upgraded that were specifically 7.0.0 and their latest is 8.0.0. For some reason I excluded these from the search. Now I know.

Piedone (approved):

Psichorex:
https://github.com/Lombiq/NodeJs-Extensions/actions/runs/7887385794
This happens when trying to publish to NuGet.
As far as I am concerned collection literals are not a preview feature. I didn't find a corresponding part of the https://lombiq.atlassian.net/wiki/spaces/DEV/pages/786857987/Managing+releases+of+open-source+projects whether what to do if a .NET upgrade happens for instance.
Do you have any idea or suggestions how to overcome this?

Piedone:
NE targets `netstandard2.0` and should keep doing that for compatibility with .NET Framework projects. You can't use new C# features there.

Psichorex:
> NE targets `netstandard2.0` and should keep doing that for compatibility with .NET Framework projects. You can't use new C# features there.

That's what I was going to ask this moment. Whether NE actually brings the langVersion from the global setting. 
BUT this error is not present locally. Which means if I don't use collection literals it will induce warnings in the solution. 
So shall we supress these warnings of manipulate that rule?

Piedone:
You can also specify `<langVersion>` just in the csproj of NE, with a comment on why that's necessary, to avoid the warnings.

Psichorex:
The default `<LangVersion>` for netstandard2.0 is C# 7.3. But that also doesn't include file scoped namespaces. So what language version shall we use there? Btw namespaces were introduced in C# 10 which requires .NET 6 so I don't even understand how is that usable in a netstandard2.0 project and the CI didn't warn us before about that one.

Piedone:
Please use what works with netstandard2.0. But I don't think you need to do a release here, since only the sample is updated, and this project doesn't depend on OC.

Psichorex:
> Please use what works with netstandard2.0. But I don't think you need to do a release here, since only the sample is updated, and this project doesn't depend on OC.

I adjusted the syntaxes used in `Lombiq.NodeJs.Extensions` you can see the changes in the the PR:https://github.com/Lombiq/NodeJs-Extensions/pull/91/files#diff-9cf2858bf3bd132496ec78748544fe8bb1e628b8244b6e65be188fa6f774b5de
But as you said in this case I won't actually do a release here because only the `.Tests` project is modified by the .NET upgrade and it is OC independent.

Piedone:
Thanks!

Psichorex:
CI Tests are failing due to this:
image

I will increase the timeout now but I have a feeling that this shouldn't be increased.
Does .NET 8 have maybe anything to do with increased timeout demands? We need them both on windows and ubuntu now on the larger runners. I am not an expert in this field I haven't yet looked into timeout values and why we need them. All I know is that you have mentioned it should be less than `720000` on larger runners.

Piedone:
That timeout is already large and shouldn't be increased without adding any further tests. Most possibly a test is flaky or downright failing.

Psichorex:
> That timeout is already large and shouldn't be increased without adding any further tests. Most possibly a test is flaky or downright failing.

Maybe related: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/694 but here nothing was touched that should cause such a case. It was derived from latest stable dev and only a PS1 script was changed.

Psichorex:
Also this: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/7905683913?pr=673
I see that this was recently fixed but it seems that the fix doesn't cover all cases. Correspoding issue is OSOE-771https://lombiq.atlassian.net/jira/software/c/projects/OSOE/boards/36?selectedIssue=OSOE-771

Piedone:
So, it's indeed a failing test. You need to fix that (looks like an incompatibility that 1.8 brought) and revert the timeout values.

test/Lombiq.OSOCE.Tests.UI/Tests/ModuleTests/BehaviorWalkthroughsTests.cs
```diff
@@ -41,8 +41,7 @@ public BehaviorWalkthroughsTests(ITestOutputHelper testOutputHelper)
     private static bool IsValidLogEntry(LogEntry logEntry) =>
         OrchardCoreUITestExecutorConfiguration.IsValidBrowserLogEntry(logEntry) &&
         // See https://github.com/OrchardCMS/OrchardCore/issues/15301.
-        !(logEntry.Message.ContainsOrdinalIgnoreCase("/OrchardCore.Resources/Scripts/jquery.js?v=") &&
```
Piedone, line 44: This should be more specific than that, since this'd ignore all jQuery exceptions.

Psichorex: I just merged this code into my branch and my CI started failing. 
I deleted `3138:6 Uncaught` because my CI dropped `3137:6 Uncaught` so it didn't do anything.

Piedone: What's the full error message currently and why? Investigate it.

Psichorex: The error message is the same for what you have added this log filter in OSOE-771. The only difference is that filter is not filtering enough on my branch and it failed. I probably also have some problems with the `dotnet-dump tool` it seems that it also has a .NET version tied versioning so I updated that aswell. Now I am waiting for the CI.

Psichorex: image
This was the cause of the failure by the way. Which in my opinion should have been removed by the filter.

Piedone: This is the same error message as under OC 1.7, the string you removed is right there at the top.

I trust you to handle this from this point, so I'll only get back on the next review.

Psichorex:
Do you know how to increase the test case timeout? 
https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/7914125012/attempts/1#ra3eb2a264s18
Walkthorugh tests are timing out and I am not getting a proper error why is it actually failing.
For me locally the test passes at the first run.

Piedone:
This is a different timeout, some browser command doesn't complete, i. e. the test completes but fails. See the stack trace in the error message and download the failure dump.

Psichorex:
> This is a different timeout, some browser command doesn't complete, i. e. the test completes but fails. See the stack trace in the error message and download the failure dump.

The part that always fails is somewhere at the `Deployment plan` section.
I now went thorugh the logs and found the following:
image
This part is recursively appearing until the test times out.
And by the screenshot it uploads it seems that the popup modal is appearing in such a way that it is covering the select button of the OC modal's card element thus I can't click on it gets stuck?

Piedone:
Then, apparently that UI changed in 1.8 and you have to change the selector the Walkthroughs module uses to display the tooltip not to hide it.

Psichorex:
> Then, apparently that UI changed in 1.8 and you have to change the selector the Walkthroughs module uses to display the tooltip not to hide it.

The main problem is that he javascript that the module uses to render that tooltip is fine. Locally it is working as it should be. 
See:

`javascript`:
image
Local `data-popper-placement="top"`:
image
### CI LOGS
`data-popper-placement="bottom"`
image
image

At this point I have no clue why does the CI env change the render position from top to bottom as the script clearly says it should be top. 
I checked the button element on both CI and Local and they are exactly the same.
Local:
image
CI:
image

Psichorex:
I changed it from top to right so that window size won't really affect it and it works now.

NuGetTest/src/Lombiq.OSOCE.NuGet.Web/Lombiq.OSOCE.NuGet.Web.csproj
```diff
@@ -69,4 +69,8 @@
     <ProjectReference Include="..\Themes\Lombiq.OSOCE.NuGet.TestTheme\Lombiq.OSOCE.NuGet.TestTheme.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Update="Lombiq.Analyzers.OrchardCore" Version="4.0.1-alpha.0.osoe-751" />
+  </ItemGroup>
```
Piedone: This should be done in the Directory.Build.props file, not every project.

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Helpers/AssertBrowserLogHelpers.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Threading.Tasks;
+
+namespace Lombiq.OSOCE.NuGet.Tests.UI.Helpers;
+public static class AssertHtmlAndBrowserErrorsHelper
+{
+    public static readonly Action<IEnumerable<LogEntry>> AssertBrowserLogIsEmpty =
+        logEntries => logEntries.Where(
+            logEntry => !logEntry.Message.ContainsOrdinalIgnoreCase("Prefer to use the native <button> element"));
+
+    public static readonly Func<HtmlValidationResult, Task> AssertHtmlErrorsAreEmpty = async errors =>
```
Piedone: None of them assert the log is empty, use a more descriptive name.

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Tests/ModuleTests/BehaviorChartJsTests.cs
Piedone, line 1: I don't see the reason to do these formatting changes is any of the test classes.

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Tests/ModuleTests/BehaviorWalkthroughsTests.cs
```diff
@@ -32,17 +29,16 @@ public BehaviorWalkthroughsTests(ITestOutputHelper testOutputHelper)
...
-        OrchardCoreUITestExecutorConfiguration.IsValidBrowserLogEntry(logEntry) &&
-        // See https://github.com/OrchardCMS/OrchardCore/issues/15301.
-        !(logEntry.Message.ContainsOrdinalIgnoreCase("/OrchardCore.Resources/Scripts/jquery.js?v=") &&
-            logEntry.Message.ContainsOrdinalIgnoreCase("3128:6 Uncaught")) &&
-        // See https://github.com/OrchardCMS/OrchardCore/issues/14598. This error has multiple variations, so targeting
-        // the lowest common denominator with the file name.
-        !logEntry.Message.ContainsOrdinalIgnoreCase("/monaco/IStandaloneEditorConstructionOptions.json");
+                    OrchardCoreUITestExecutorConfiguration.AssertBrowserLogIsEmpty(filteredLogEntry);
+                };
+            });
```
Piedone: Why rearrange this? Note that the custom message in the assertion (second parameter of `ShouldNotContain()`) is now also removed, making debugging harder. Same for this test in the root solution.

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Tests/ModuleTests/BehaviorWalkthroughsTests.cs
```diff
@@ -32,17 +29,16 @@ public BehaviorWalkthroughsTests(ITestOutputHelper testOutputHelper)
...
                 // Once the linked issues are fixed, the custom browser log assertion can be removed completely.
-                configuration.AssertBrowserLog = logEntries => logEntries.ShouldNotContain(
-                    logEntry => IsValidLogEntry(logEntry),
-                    logEntries.Where(IsValidLogEntry).ToFormattedString());
-            });
+                configuration.AssertBrowserLog = logEntries =>
+                {
+                    var filteredLogEntry = logEntries.Where(logEntry =>
+                        // See https://github.com/OrchardCMS/OrchardCore/issues/15301.
+                        !logEntry.Message.ContainsOrdinalIgnoreCase("/OrchardCore.Resources/Scripts/jquery.js?v=") &&
```
Piedone: As I said under https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/673#discussion_r1489947345, this is not specific enough. This would ignore all jQuery errors. Same for this test in the root solution.

Piedone (changes requested):

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Tests/ModuleTests/BehaviorWalkthroughsTests.cs
Psichorex: Because the previous version of this code didn't work for me. But I will re-test this.

NuGetTest/src/Lombiq.OSOCE.NuGet.Web/Lombiq.OSOCE.NuGet.Web.csproj
Psichorex: My bad. I used the NuGet manager and it added it into everywhere after upgrading.

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Tests/ModuleTests/BehaviorChartJsTests.cs
Psichorex: These are most probably CTRL + K + D changes.

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Tests/ModuleTests/BehaviorWalkthroughsTests.cs
Psichorex: Okay so the problem was that `3138:6` is too specific and is subject to change. So using only `Uncaught` will solve this.

NuGetTest/test/Lombiq.OSOCE.NuGet.Tests.UI/Tests/ModuleTests/BehaviorChartJsTests.cs
Piedone: It's not happening for me, and they shouldn't anyway. Please revert all of them.

Piedone (approved):

PR #724, opened 2024-03-15
OSOE-819: User source generator to determine version

AydinE:
[OSOE-819](https://lombiq.atlassian.net/browse/OSOE-819)
Get Vue version directly from package.json using the new source generators project - see: https://github.com/Lombiq/Helpful-Libraries/issues/238

[OSOE-819]: https://lombiq.atlassian.net/browse/OSOE-819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sarahelsaig:
Please replace [this](https://github.com/Lombiq/Orchard-Base-Theme/blob/dev/Lombiq.BaseTheme/ResourceManagementOptionsConfiguration.cs#L16) as well. I think that's the only const version left in OSOCE.

AydinE:
> Please replace [this](https://github.com/Lombiq/Orchard-Base-Theme/blob/dev/Lombiq.BaseTheme/ResourceManagementOptionsConfiguration.cs#L16) as well. I think that's the only const version left in OSOCE.

Added this one also!

PR #787, opened 2024-06-05
OSOE-400: Auto-generate next version number when publishing NuGet packages

AydinE:
[OSOE-400](https://lombiq.atlassian.net/browse/OSOE-400)

[OSOE-400]: https://lombiq.atlassian.net/browse/OSOE-400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Get-Solution-Projects.ps1
sarahelsaig, line 1: Please rename this file to _Get-SolutionProject.ps1_ to follow the Powershell cmdlet naming conventions
([the noun part should be a single PascalCase word](https://learn.microsoft.com/en-us/powershell/scripting/developer/cmdlet/strongly-encouraged-development-guidelines?view=powershell-7.4#use-pascal-case-for-cmdlet-names-sd02) and [it should be singular instead of plural](https://learn.microsoft.com/en-us/powershell/scripting/developer/cmdlet/strongly-encouraged-development-guidelines?view=powershell-7.4#use-singular-parameter-names)).

Get-Solution-Projects.ps1
```diff
@@ -0,0 +1,42 @@
...
+
+if (-not $solutionFile)
+{
+    throw 'No solution file found in the specified path.'
+}
+
+if (-not $NewVersion)
+{
+    throw 'No new version specified.'
+}
```
sarahelsaig: Mark the parameter mandatory instead of manually checking with `[Parameter(Mandatory=$true)]`([see docs](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_parameters?view=powershell-7.4#parameter-attribute))

Get-Solution-Projects.ps1
```diff
@@ -0,0 +1,42 @@
...
+    # Check if the line starts with "Project"
+    if ($line -match '^Project\("' -and $line -like '*.csproj*')
+    {
+        # Extract the project name from the line
+        $projectName = ($line -split ',')[0].Trim('"').Split('=')[1].Trim(' "')
+
+        # Add the project name to the array
+        $projectNames += $projectName
+    }
+}
```
sarahelsaig: This is not necessary, you can use the dotnet CLI like `dotnet sln <SLN_FILE> list`.

Update-References.ps1
```diff
@@ -0,0 +1,39 @@
+# Call like this: .\Update-References.ps1 -ProjectToFind "ProjectName" -NewVersion "1.0.0"
+param(
+    [string]$ProjectToFind, # Name of the project to check reference for
+    [string]$NewVersion # New version number to set
+)
```
sarahelsaig: Use [comment based help](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comment_based_help?view=powershell-7.4) instead of regular comments. It's more structured and `Get-Help` will display it as well.

Update-References.ps1
```diff
@@ -0,0 +1,39 @@
...
+
+# Get all .csproj files recursively
+$projectFiles = Get-ChildItem -Path .\ -Recurse -Filter '*.csproj'
+
+# Loop through each project file to see if it has a reference to the project we are looking for
+ForEach ($projectFile in $projectFiles)
```
sarahelsaig: You use `foreach` all lower case everywhere else ([and so does the documentation](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_foreach?view=powershell-7.4)). Let's keep it consistent.

```suggestion
foreach ($projectFile in $projectFiles)
```

Update-References.ps1
```diff
@@ -0,0 +1,39 @@
+# Call like this: .\Update-References.ps1 -ProjectToFind "ProjectName" -NewVersion "1.0.0"
+param(
+    [string]$ProjectToFind, # Name of the project to check reference for
+    [string]$NewVersion # New version number to set
+)
+
+# Get all .csproj files recursively
+$projectFiles = Get-ChildItem -Path .\ -Recurse -Filter '*.csproj'
```
sarahelsaig: Path is assumed to be `.` by default so that's not necessary.

```suggestion
$projectFiles = Get-ChildItem -Recurse -Filter '*.csproj'
```

Update-References.ps1
```diff
@@ -0,0 +1,39 @@
...
+    # Set PreserveWhitespace to true
+    $projectXml.PreserveWhitespace = $true
+
+    # Load the XML file
+    $projectXml.Load($projectFile)
+
+    # Find the ProjectReference nodes
+    $projectReferences = $projectXml.Project.ItemGroup.PackageReference
+
+    foreach ($projectReference in $projectReferences)
```
sarahelsaig: Since this is for `<PackageReference>` rename these to `$packageReference` and `$packageReferences` respectively.

Update-References.ps1
```diff
@@ -0,0 +1,39 @@
...
+    {
+        # Check if the Include attribute contains the project name
+        if ($projectReference.Include -like "*$ProjectToFind*")
+        {
+            if ($projectReference.Version)
```
sarahelsaig: Merge these two nested conditions with `-and`.

VNext.ps1
```diff
@@ -0,0 +1,124 @@
+# Call like this: .\VNext.ps1 -Path '\Path\to\project' -PreRelease $True -Issue '1234' -UpdateReferences $true
```
sarahelsaig: You use both `$True` and `$true` across the PR. Please replace all `$True` with `$true` for consistency.

VNext.ps1
```diff
@@ -0,0 +1,124 @@
...
+# or like this: .\VNext.ps1 -Path '\Path\to\project' -Type 'major' -UpdateReferences $true
```
sarahelsaig: Replace these with comment based help as well.

VNext.ps1
```diff
@@ -0,0 +1,124 @@
...
+    [string]$Issue,
+    [bool]$PreRelease = $false,
+    [bool]$UpdateReferences = $false,
+    [bool]$Interactive = $true
+)
+
+$origin = Get-Location
+$newVersion = ''
+
+if ($PreRelease -eq $false -and -not $PSBoundParameters.ContainsKey('Type'))
```
sarahelsaig: Is there a reason why you used `$PreRelease -eq $false` instead of `-not $PreRelease`?

VNext.ps1
```diff
@@ -0,0 +1,124 @@
+# Call like this: .\VNext.ps1 -Path '\Path\to\project' -PreRelease $True -Issue '1234' -UpdateReferences $true
+# or like this: .\VNext.ps1 -Path '\Path\to\project' -Type 'major' -UpdateReferences $true
+param(
+    [string]$Path,
+    [string]$Type,
+    [string]$Issue,
+    [bool]$PreRelease = $false,
+    [bool]$UpdateReferences = $false,
+    [bool]$Interactive = $true
```
sarahelsaig: Please turn these into [`[switch]` parameters](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_parameters?view=powershell-7.4#switch-parameters) and replace `[bool]$Interactive = $true` with `[switch]$NonInteractive` so that can be false by default as well. It's best to use switch instead of bool, because then you can invoke it like
```pwsh
.\VNext.ps1 -Path '\Path\to\project' -Type 'major' -UpdateReferences
```
without having to pass the `$true`. It's how the built-in cmdlets behave so it's more familiar to the user.

VNext.ps1
```diff
@@ -0,0 +1,124 @@
...
+
+$origin = Get-Location
+$newVersion = ''
+
+if ($PreRelease -eq $false -and -not $PSBoundParameters.ContainsKey('Type'))
+{
+    throw 'Must set a version Type (major, minor, patch)'
+    if ($Type -ne 'major' -and $Type -ne 'minor' -and $Type -ne 'patch')
+    {
+        throw 'Invalid version Type must be major, minor, or patch'
```
sarahelsaig: Put the values in quotes.

VNext.ps1
```diff
@@ -0,0 +1,124 @@
...
+    }
+}
+
+if ($PreRelease -eq $true -and -not $PSBoundParameters.ContainsKey('Issue'))
```
sarahelsaig:
```suggestion
if ($PreRelease -and -not $PSBoundParameters.ContainsKey('Issue'))
```

VNext.ps1
```diff
@@ -0,0 +1,124 @@
...
+    throw 'Pre-releases must have an Issue number'
+}
+
+if ($PSBoundParameters.ContainsKey('Path'))
+{
+    Set-Location $Path
+}
+
+#Get the latest release tags, we get the last 20 tags so we can determine the next version or pre-release version
+$latestTags = git for-each-ref --sort=-creatordate --count=20 --format '%(refname:short)' refs/tags
```
sarahelsaig: Why is the number 20 necessary here? It works for me without it just as well. Won't this break if there were more than  20 prerelease versions?

VNext.ps1
```diff
@@ -0,0 +1,124 @@
...
+    Set-Location $Path
+}
+
+# Get the latest release tags, we get the last 20 tags so we can determine the next version or pre-release version
+$latestTags = git for-each-ref --sort=-creatordate --count=20 --format '%(refname:short)' refs/tags
+
+# Check tags for a valid version number and try to determine the next version
+try
+{
+    if ($latestTags[0] -match 'v\d+\.\d+\.\d+')
```
sarahelsaig: Check from string start.
```suggestion
    if ($latestTags[0] -match '^v\d+\.\d+\.\d+')
```

sarahelsaig: Aren't you missing a foreach here? It seems to me that now it's only checking the first tag.

VNext.ps1
```diff
@@ -0,0 +1,124 @@
...
+    {
+        $latestVersion = $matches[0]
+        $version = $latestVersion -split '\.'
+        $major = [int]$version[0].Substring(1)
+        $minor = [int]$version[1]
+        $patch = [int]$version[2].Substring(0, 1)
```
sarahelsaig: It's safer to convert it into a `Version` instance with `[Version]$matches[0].Substring(1)`.

VNext.ps1
```diff
@@ -0,0 +1,124 @@
...
+            }
+            'patch'
+            {
+                $patch++
+            }
+        }
+
+        $newVersion = "v$major.$minor.$patch"
+
+        if ($PreRelease -eq $true)
```
sarahelsaig:
```suggestion
        if ($PreRelease)
```

VNext.ps1
```diff
@@ -0,0 +1,124 @@
...
+        {
+            # Check if there were any pre-release tags for this Issue already in latestTags array
+            $preReleaseTags = $latestTags | Where-Object { $PSItem -match ([regex]::escape($Issue)) }
```
sarahelsaig:
```suggestion
            $preReleaseTags = $latestTags | Where-Object { $PSItem -match ([regex]::Escape($Issue)) }
```

VNext.ps1
```diff
@@ -0,0 +1,124 @@
...
+            if ($preReleaseTags.Count -gt 0)
+            {
+                $latestPreRelease = $preReleaseTags[0]
```
sarahelsaig: If there was only one prerelease for an issue then `$preReleaseTags` will be `[string]` and `$latestPreRelease` will be just `"v"`:
image

VNext.ps1
```diff
@@ -0,0 +1,124 @@
...
+                exit
+            }
+        }
+
+        git tag $newVersion
+        git push origin tag $newVersion
+    }
+    else
+    {
+        Write-Output 'No valid tags found'
```
sarahelsaig:
```suggestion
        Write-Output 'No valid tags were found.'
```

Releasing.md
```diff
@@ -0,0 +1,90 @@
+﻿# Releasing a new version
+
+For releasing a new version of any project in this repository to NuGet there are a few handy scripts that you can use.
```
sarahelsaig:
```suggestion
We have a few handy scripts you can use to release a new NuGet package version of any project in this repository.
```

Releasing.md
```diff
@@ -0,0 +1,90 @@
...
+- `PreRelease`: Boolean flag to indicate if this is a pre-release.
+- `Issue`: The issue number that this release is related to, is only used if it is a pre-release.
+- `UpdateReferences`: Boolean flag to indicate if the references should be updated to this new version in the other projects in the repository.
+- `Interactive`: Boolean flag to indicate if the script should ask for confirmation before pushing changes to git - defaults to true.
+
+## VNext - Usage and examples
+
+### Releasing a new major version
+
+```shell
```
sarahelsaig: `shell` refers to Unix _.sh_ scripts. Please replace with `pwsh` across the document.

Releasing.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+### Example release of a new major version
+
+1. Open a PowerShell terminal.
+2. Navigate to the root of the repository.
+3. Run the script with the necessary parameters. For example to release a new major version of the `Lombiq.HelpfulLibraries` project and update the references in other projects, run the following command
+
+    ```shell
+    .\VNext.ps1 -path .\src\Libraries\Lombiq.HelpfulLibraries\ -type "major" -commitChanges $true
```
sarahelsaig: Make sure all the parameter names have the correct casing, e.g. `-Path`, `-Type`, etc. (same below)

VNext.ps1
sarahelsaig, line 1: Consider renaming this file to `Publish-VNext` to match the code style of using [an approved verb](https://learn.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7.4#data-verbs)

sarahelsaig (changes requested):

VNext.ps1
AydinE: For this section I am indeed only checking the first tag, it's only needed to look at the rest of the tags in case of a pre-release (just in case there were other releases of the same package in between pre-releases for the same issue).
But let me know if I am missing some edge cases here maybe

AydinE:
I believe I have addressed all of the points, since implementing your suggestion of using `dotnet sln list` which is obviously much cleaner I have decided the `Get-Solution-Projects` script was no longer needed so this is now removed.

VNext.ps1
sarahelsaig: @AydinE you didn't respond to this.

AydinE: @sarahelsaig Sorry must have missed this one, I only have 20 as an arbitrary number really. This could have been any number and is really only in some case where maybe a (pre-)release of another issue was published in between 2 versions of the current issue. I am not sure how common that would be or how much of an edge case that is but I built it like that just in case. 

So in case we were both working on issues in the same repo and you released v1.1.1-alpha.1.osoe-123 and I published v1.1.1-alpha.1.osoe-321 and then you wanted to do v1.1.1-alpha.2.osoe-123 after that it would not work without this. But perhaps this is overthinking it?

sarahelsaig: > Sorry must have missed this one

If you put an emoji reaction (like :+1: or :eyes:) on a comment you've already addressed then it's easy to spot if there is anything left out.

> This could have been any number and is really only in some case where maybe a (pre-)release of another issue was published in between 2 versions of the current issue.

But why constrain it at all? If you remove the `--count=20` argument it still works and we will never have so many tags where fetching the compete list could cause an issue.

> So in case we were both working on issues in the same repo and you released v1.1.1-alpha.1.[osoe-123](https://lombiq.atlassian.net/browse/OSOE-123) and I published v1.1.1-alpha.1.[osoe-321](https://lombiq.atlassian.net/browse/OSOE-321) and then you wanted to do v1.1.1-alpha.2.[osoe-123](https://lombiq.atlassian.net/browse/OSOE-123) after that it would not work without this. But perhaps this is overthinking it?

You have `--sort=-creatordate` and filter the results by `$Issue` so I don't think that can ever be a problem.

AydinE: Good call, I have removed the arbitrary count constraint now

PR #962, opened 2025-02-08
OFFI-134: Add IMAP-specific functionality to UITT

barthamark:
[OFFI-134](https://lombiq.atlassian.net/browse/OFFI-134)

[OFFI-134]: https://lombiq.atlassian.net/browse/OFFI-134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -34,4 +34,4 @@ XWidgets
 yuv
 ZXing
 imapport
-Ssl
+ssl
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -33,3 +33,5 @@ Wxga
 XWidgets
 yuv
 ZXing
+imapport
+Ssl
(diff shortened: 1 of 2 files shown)


.github/actions/spelling/allow.txt
```diff
@@ -33,3 +33,5 @@ Wxga
 XWidgets
 yuv
 ZXing
+imapport
+Ssl
```
sarahelsaig: I'm surprised that SSL is not already on the list. Anyway, unless the capitalization is significant always use all lower case, because that tells the spell checker that the pattern is case-insensitive
```suggestion
ssl
```

sarahelsaig (changes requested):

sarahelsaig (changes requested):
Please add the new submodule to OSOCE and the projects to the solution file.

barthamark:
> Please add the new submodule to OSOCE and the projects to the solution file.

If you refer to the Lombiq.EmailClient, it'll still need some preparations to make it a fully compliant Lombiq OSS module and that's planned to be in a separate issue.

sarahelsaig:
> If you refer to the Lombiq.EmailClient, it'll still need some preparations to make it a fully compliant Lombiq OSS module and that's planned to be in a separate issue.

Yep. That's fine, but have you created that follow-up issue yet? I don't see it in OSOE in Jira. If not, please create it and then set OFFI-134 as a blocker so they are linked together.

PR #1050, opened 2025-05-07
OSOE-1107: Reduce dependency on Lombiq.NodeJs.Extensions

sarahelsaig:
[OSOE-1107](https://lombiq.atlassian.net/browse/OSOE-1107)

[OSOE-1107]: https://lombiq.atlassian.net/browse/OSOE-1107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/build-and-test.yml
Piedone, line 1: I get this in VS:

{08FE1EE5-8AE5-403E-B987-C9FBFBE7C1F8}

Even though UITT doesn't use that task.

Despite this, the build succeeds, and when opening the solution again, the error is gone.

This in itself is not nice but can be OK-ish, but might show an actual problem. Isn't this caused by no project references to NPM Targets being added to consumer projects? Only props and targets files. This means that NPM Targets might not be built before its consumer projects (UITT showing here might just be a red herring caused by some VS build idiosyncrasy). I think we had the NE project references for the same reason.

src/Modules/Lombiq.OSOCE.Samples/Navigation.cs
Piedone, line 1: Thank you! This should rather be added to _Lombiq.UIKit/Showcase_, though.

.github/workflows/build-and-test.yml
Piedone, line 1: Also update the NuGetTest solution's package references with the updates you already published, as well as the changed consumer projects so we can see that NuGet publishing correctly embeds the files from the wwwroot folder.

.github/workflows/build-and-test.yml
```diff
@@ -60,6 +60,22 @@ jobs:
...
+        src/Modules/Lombiq.DataTables/Lombiq.DataTables,
+        src/Modules/Lombiq.HelpfulExtensions/Lombiq.HelpfulExtensions,
+        src/Modules/Lombiq.UIKit/Lombiq.UIKit,
+        src/Modules/Lombiq.Walkthroughs/Lombiq.Walkthroughs
+      styles-css: >-
+        src/Modules/Lombiq.Privacy/Lombiq.Privacy,
+        src/Modules/Lombiq.HelpfulExtensions/Lombiq.HelpfulExtensions,
+        src/Modules/Lombiq.UIKit/Lombiq.UIKit,
+        src/Modules/Lombiq.JsonEditor/Lombiq.JsonEditor,
+        src/Modules/Lombiq.ChartJs/Lombiq.ChartJs.Samples
```
Piedone: Also add this to the Windows workflow.

.github/workflows/build-and-test.yml
```diff
@@ -60,6 +60,22 @@ jobs:
     name: Codespell
     uses: Lombiq/GitHub-Actions/.github/workflows/codespell.yml@dev
 
+  asset-lint:
+    name: Lint Scripts, Styles, and Texts
```
Piedone: This would look better with the job being named just Lint, or Lint Assets. Then, since the workflow is named "Lint Scripts, Styles, and Texts", that'll be displayed anyway, but not twice, and we wouldn't have such a long title in the root list:

{E67D5F68-1DA3-4457-94E1-B320C444B721}


See all the other workflows too.

Piedone (changes requested):

.github/workflows/build-and-test.yml
sarahelsaig: Why? The windows workflow don't have codespell either, and this is similar in that it's only static analysis.

Piedone: Linters can behave differently, also due to line ending differences. We've seen this with .NET analyzers and PSScriptAnalyzer too.

sarahelsaig: Have any of those differences ever been anything but false positive? Seems like an unnecessary maintenance burden to me.

Piedone: The .NET ones perhaps mostly yes, but not always, the PSScriptAnalyzer were actual violations.

Since we should be able to run these linters locally too, for a quicker feedback loop, they must work equally well under Windows too, so if there are false positives or any differences between the platforms, we'll need to adapt to them.

.github/workflows/build-and-test.yml
sarahelsaig: Actually UITT [already uses that task](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/Lombiq.Tests.UI.csproj#L109-L110) in dev branch.

Regardless, I've added ProjectReference to Npm.Targets to all projects that import those targets.

Piedone: This still happens, unfortunately, but since everything else works, even the build when this happens, this must be some VS quirk.

src/Lombiq.OSOCE.Web/Lombiq.OSOCE.Web.csproj
```diff
@@ -66,4 +66,11 @@
     <PackageReference Include="OrchardCore.Logging.NLog" Version="$(OrchardCoreVersion)" />
   </ItemGroup>
 
+  <!-- Necessary to prevent package downgrade. -->
```
Piedone: Where is that coming from, exactly? Can we remove it at one point? Document it here and in NPM Targets.

Piedone (changes requested):

src/Lombiq.OSOCE.Web/Lombiq.OSOCE.Web.csproj
sarahelsaig: I'm not sure, but after merging from dev it went away, so I've removed it.

Piedone (approved):

src/Lombiq.OSOCE.Web/Lombiq.OSOCE.Web.csproj
sarahelsaig: This (`NU1107: Version conflict detected for Microsoft.Build.Tasks.Core`) showed up again during local builds, but I managed to find and [fix the underlying problem](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/1050/commits/5074232625b0c704bebda48a94f13083caa596a9). It was unrelated to this issue, instead it was caused by [this change](https://github.com/Lombiq/Hosting-Tenants/commit/1facf8b58605188a14762837ddc706ae18aa5107) in the `dev` of Lombiq.Hosting.Tenants. The project was silently importing the NuGet version of NE without anyone realizing until it caused a version conflict with the submodule version of NE.

Piedone: Eh. I guess this is necessary when OCC is mixed into the solution: https://github.com/search?q=org%3ALombiq%20LombiqNodeJsExtensionsPath&type=code Good catch, thank you.

Piedone (approved):

Piedone (approved):

PR #1243, opened 2026-02-28
NEST-644: Adding Tailwind build tools and Lucide icon picker

barthamark:
[NEST-644](https://lombiq.atlassian.net/browse/NEST-644)


Piedone (changes requested):
Follow https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions?tab=readme-ov-file#adding-a-new-extension-or-significant-new-features.

Lombiq.OSOCE.sln
Piedone, line 1: > If the project is published on NuGet, reference it from the app in the Lombiq.OSOCE.NuGet solution as well, and enable its features in the Lombiq.OSOCE.NuGet.Tests.recipe. If it has UI testing methods, also run them from Lombiq.OSOCE.NuGet.Tests.UI.

Piedone (changes requested):
And follow it fully, not just partially.

Lombiq.OSOCE.sln
barthamark: I saw this but it doesn't make sense because there's nothing to compile there. Or should I still add it to make sure the NuGet works?

Piedone: Yes, and make it used by the sample (the sample can double-use the NuGet in the NuGet solution, and project in the root solution).

barthamark: I'm still confused a little bit. I created a Lombiq.Tailwind.Targets.Samples project, a very simple .NET project with zero functionality but has an Assets/Styles.site.css and some sample file with Tailwind classes and since it references the target files the compiled .css file will appear in wwwroot. OK so I added this to OSOCE. It works. But what exactly should I test or do now? I can't imagine the test case here that we should automate.

Piedone: Test that the correct CSS is generated. Not to test the Tailwind compiler itself, but rather, to check that it actually has run for the template and contains code for the classes in the template. This you can do from a unit, not UI test too, by reading the CSS file and asserting on its content.

barthamark: Should it be an Orchard Core theme or just a simple library and somehow figure out how to test it from a unit test?

Piedone: A theme so it also serves as an example.

Piedone (approved):

Piedone (approved):

Piedone (approved):

